PR #36, opened 2024-03-22
add stopping rule if # incorrect responses exceeded

kachergis:


Zio-4 (commented):
It always counts the practice trials towards the number of incorrect.

PR #40, opened 2024-03-28
Fix - experiment ending early from corpus row count mismatch

Zio-4:
- Changing variable name maxStimulusTrials to totalTrials 
- Now accounts for all trials minus the amount of practice trials NOT given (Ex. 100 rows in corpus = 100 trials, maxPracticeTrials = 2, total trials will be 92).
- Always includes instructions. I assume that we will always want to give the amount of instructions listed in the corpus.


Zio-4:
@kachergis Going to merge so Tonya can continue working but tagging so you can see the changes

PR #43, opened 2024-04-08
Adding styling from Figma, pulling out code, updating data saved to db

Zio-4:
- Adding styling to stimulus trials from Figma
- Pulled out on_load code into a function
- Abstracted corsi block trials into a function that returns the trials, similar to afcStimulus.
- Updated data saved to Firestore. No longer saving save_trial flag and internal node id from jsPsych.

### DIFF
--- a/task-launcher/src/styles/task.scss
+++ b/task-launcher/src/styles/task.scss
@@ -630,16 +630,22 @@
 
       .jspsych-corsi-overide {
         display: grid;
-        gap: 5px; /* adjust as needed for spacing between blocks */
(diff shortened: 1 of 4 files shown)


kachergis (commented):
styling changes look great! just want to note that we still need the instructions (see video in `core-tasks/assets/memory-game/docs/`) and to implement the non-practice blocks (without feedback), including forward and backward phases with adaptively increasing grid sizes as specified below:
Screenshot 2024-04-09 at 5 10 37 PM

PR #55, opened 2024-04-17
more responsive slider layout

mikabr:


kachergis:
thank you!

kachergis (commented):
looks good -- thank you! agree that we should do a CSS clean-up after this pilot (is @natling interested??)

PR #62, opened 2024-04-24
Saving trials if flag turned on after a trial is finished

asengupta3:
Screenshot 2024-04-26 at 1 32 26 AM

### DIFF
--- a/task-launcher/src/tasks/shared/helpers/initTrialSaving.js
+++ b/task-launcher/src/tasks/shared/helpers/initTrialSaving.js
@@ -17,7 +17,10 @@ export const initTrialSaving = (config) => {
       code.apply(fn, arguments);
     };
 
-  jsPsych.opts.on_finish = extend(jsPsych.opts.on_finish, () => {
+  jsPsych.opts.on_finish = extend(jsPsych.opts.on_finish, (data) => {
+    if (data.save_trial) {
+      config.firekit.writeTrial(data);
+    }
     config.firekit.finishRun();
   });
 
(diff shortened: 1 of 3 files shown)


task-launcher/src/tasks/shared/helpers/initTrialSaving.js
```diff
@@ -17,7 +17,10 @@ export const initTrialSaving = (config) => {
       code.apply(fn, arguments);
     };
 
-  jsPsych.opts.on_finish = extend(jsPsych.opts.on_finish, () => {
+  jsPsych.opts.on_finish = extend(jsPsych.opts.on_finish, (data) => {
+    if (data.save_trial) {
+      config.firekit.writeTrial(data);
+    }
     config.firekit.finishRun();
```
Zio-4, line 24: The logic here actually doesn't need any change. In afcStimulus.js, in the shared folder, in the on_finish function, we should check if the final data includes the corpus trial type, and if not, add it to the data object. The trial type will come from the current stimulus object. It looks like there is a trialType property being saved but perhaps it is being overwritten by jsPsych after our code.

Zio-4 (changes requested):

asengupta3:
@Zio-4 I am seeing it now in the DB (please see the screenshot in the description), but not trials have it, I wonder why. The ones that don't have it looks like this:
Screenshot 2024-04-26 at 1 34 07 AM

Zio-4 (changes requested):
Looks good. Could you remove the logs? We don't have any blocks currently so after that you can merge it. Thanks!

Zio-4:
Responding to your comment: It seems to be instruction trials don't save the trial type. I tested and noticed this on Math. This is fine. In fact, we don't want to write any instruction trials to the DB so we can add that to the backlog.

PR #66, opened 2024-04-27
Heart and Flowers refactor and fix various bugs

anabonne:
- Added a pulse animation on the "valid-answer" button for instruction-practice trials and feedback trials
- Fixed non-flower expected answer being flipped in data recorded from stimulus trials
- Refactored the trials instantiations to be dynamic and done upon calling builder methods from the main buildHeartsAndFlowersTimeline() method in timeline.js
- Removed function wrapper around static assets
- Converted instruction and feedback trials to Audio
- Introduced a hard coded timeline config
- added feedback trials after each practice trial
- Fixed self-inflicted exception bug at the beginning of second mixed set of trials


task-launcher/serve/serve.js
```diff
@@ -14,6 +14,8 @@ const taskName = urlParams.get('task') ?? 'egma-math';
 const corpus = urlParams.get('corpus');
 const buttonLayout = urlParams.get('buttonLayout');
 const numOfPracticeTrials = urlParams.get('practiceTrials');
+//TODO: remove default values: Curious why we would use default values here in case we cannot parse these int,
```
anabonne, line 17: I got this wrong it's the radix not a default value!

task-launcher/src/styles/task.scss
```diff
@@ -625,8 +625,17 @@ $breakpoint-large-desktop: 1440px;
       .btn-layout-hf {
```
anabonne, line 625: This is the only change that impacts a file common to the other tasks. 
As indicated by the suffix '-hf' it is to my knowledge not used outside of the Hearts and Flowers task code.

kachergis (commented):
- should make the instruction image size (e.g., bear, skating, etc.) responsive, to scale with window width/height (currently "OK" button can be cut off at bottom)
- SP audio feedback is "Buen trabajo!" and text is "Bien hecho!" (prob not dev fault)
- may need to set a min-width for the stimulus box (will consult Michael Sulik)
- we will want to change the look-and-feel of the buttons and background to be more like our other tasks (see the figma design)
- also, Phil generally likes to remove all console.log statements

PR #69, opened 2024-04-29
 HaF: stimulus presentation time

anabonne:
just hardcoding stimulus time to configs from timeline.js


Zio-4:
For some reason, these changes break the task when you respond with arrow keys. Could you look into this?

Zio-4 (approved):

PR #73, opened 2024-04-30
Styling changes and moving to templates

asengupta3:


Zio-4:
@asengupta3 Is this ready for review? If not could you convert it to a draft? I didn't realize we could do that but we should use that feature.

task-launcher/src/tasks/hearts-and-flowers/trials/stimulus.js
```diff
@@ -22,26 +28,11 @@ export function stimulus(isPractice = false, stage, stimulusDuration, postTrialG
...
-                    </div>
-                </div>`;
+      return getLayoutTemplate(
+        null,
+        mediaAssets.images[jsPsych.timelineVariable('stimulus')],
+        jsPsych.timelineVariable('position') <= 0.5
+      );
     },
     stimulus_duration: stimulusDuration,
     post_trial_gap: isPractice? null : postTrialGap,
```
asengupta3, line 38: @anabonne the `stimulus_duration` is causing the container to disappear after sometime. Is that the expected behavior and the `post_trial_gap` is causing a flicker on the screen as well. @kachergis @Zio-4  thoughts?

kachergis: as discussed, good to not have the container disappear -- thank you!

kachergis (approved):
looks great!

task-launcher/src/tasks/hearts-and-flowers/trials/stimulus.js
anabonne: Yep we will need to make only the heart or flower disappear. For now we can just remove any timeouts on the stimulus.

PR #75, opened 2024-05-01
HaF: Last minute changes for today / this week

anabonne:
Reapplying the interStimulus interval value but this time to the fixation trial length (instead of a white screen via post_trial-gap)


kachergis (approved):
Looks great -- thanks Adrien!

Minor future ToDos:
probably remove the repetitive text from these instruction slides (could have first sentence at top, second at bottom -- or just both up top. don't need repeat sentence)
Screenshot 2024-05-01 at 1 38 51 PM

Screenshot 2024-05-01 at 1 39 02 PM

PR #79, opened 2024-05-02
HaF: implement practice trial shortcut

anabonne:
correctPracticeTrial value is read from admin config defined here (as of now it's 2, 2, and 3 for Hearts, Flowers, and Mixed practice blocks respectively): https://github.com/levante-framework/core-tasks/blob/044690ebfd37b6d405a233ceeb9aa13f32fb10bf/task-launcher/src/tasks/hearts-and-flowers/timeline.js#L54-L85

### DIFF
--- a/task-launcher/src/tasks/hearts-and-flowers/timeline.js
+++ b/task-launcher/src/tasks/hearts-and-flowers/timeline.js
@@ -83,29 +83,6 @@ export default function buildHeartsAndFlowersTimeline(config, mediaAssets) {
     },
   };
 
-  // introductionTrials, ifNotFullscreen,
-  // const timeline = [
-  //   preloadTrials,
-  //   initialTimeline,
-  //   introduction,
-  //   heartInstructions,
-  //   heartPracticeBlock1,
-  //   heartPracticeBlock2,
-  //   timeToPractice,
-  //   heartPracticeTimeline,
-  //   heartPostPracticeBlock,
-  //   heartTimeline,
-  //   flowerInstructionsBlock,
-  //   flowerPracticeTimeline,
-  //   flowerPostPracticeBlock,
-  //   flowerTimeline,
-  //   heartsAndFlowersInstructionsBlock,
-  //   heartsAndFlowersPracticeTimeline,
-  //   heartsAndFlowersPostPracticeBlock,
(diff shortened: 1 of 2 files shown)


task-launcher/src/tasks/hearts-and-flowers/timeline.js
```diff
@@ -329,15 +323,30 @@ function getMixedPracticeSection(adminConfig) {
...
+  let practiceWinStreakCount = 0;
+  function onTrialFinishTimelineCallback(data) {
+    practiceWinStreakCount = data.correct ? practiceWinStreakCount+1 : 0;
+    console.log(`Practice block win streak=${practiceWinStreakCount}`);
+    if (practiceWinStreakCount >= adminConfig.correctPracticeTrial) {
+      console.log(`Ending practice block early: win streak=${practiceWinStreakCount}`);
+      jsPsych.endCurrentTimeline();
+    }
+  };
+
```
Zio-4, line 336: Why do we need this block defined twice? Also can you remove the logs please.

anabonne: The adminConfig object is a different one and the logic is used only twice. I thought about introducing a function "getOnTrialFinishTimelineCallback(adminConfig)" that both mixed and non-mixed practice logic would call, but doing so would hide the implementation details (particularly the fact that the callback takes a parameter). Ideally we may want to create a wrapper class/object to abstract away this quirky behavior which need to be scoped to the timeline level rather than trial level.

anabonne: I may be wrong on the approach though. It's just that so far I could not see a good way to implement behaviors that span a particular nested timeline. You can pass a on_finish mapping here (as commented above circa line 260 - 250) but it gets completely overridden if you also pass one with your "trial" object.

anabonne: I think the "jsPsych way" is either to leverage the on_trial_finish callback at the experiment level or to use [conditional trials (by passing a conditional function)](https://www.jspsych.org/6.3/overview/timeline/#conditional-timelines). But both method still get you in a bit of an awkward situation (where you are either scoped to the experiment or the trial but need to apply logic on a specific nested timeline).

PR #81, opened 2024-05-03
Staggered Buttons for ToM

asengupta3:


task-launcher/src/tasks/shared/helpers/prepareChoices.js
```diff
@@ -4,7 +4,7 @@ import { mediaAssets } from '../../..';
 
 export const prepareChoices = (target, distractors, randomizeOrder = true) => {
   let choices;
-  if (distractors.includes(target)) {
+  if (!target || distractors.includes(target)) { // If target is not present, don't add to options
```
asengupta3, line 7: For audio inputs, empty target was being added which caused 3 buttons to render instead of 2 etc

kachergis (approved):
theory of mind done -- great work!

PR #82, opened 2024-05-03
Adding a centered div as a parent for haf to fix the instruction page image positioning

asengupta3:


kachergis (approved):
great, thanks!

PR #85, opened 2024-05-04
Same different

Zio-4:


kachergis (approved):
awesome, looks great! we'll want to change spacing of shapes on "here are two shapes" trials to more closely match the "here's a new shape" trials, and make sure we're scoring correctly, but looks great for testing!

PR #91, opened 2024-05-09
Hide keyhelpers and add maxtime condition

kachergis:
should test a little more.
tempted to also remove 
`afcCondtional()` from afcStimulus.js (it's unused anywhere -- what is/was it for?), and merge `afcStimulus()` with `afcStimulusWithTimeoutCondition()` -- any reason not to?


Zio-4 (approved):
Removed the old afcStimulusWithConditional. It had old practice logic that we no longer use. Changed arrow key emojis in slider trial to import from shared.

PR #94, opened 2024-05-10
Extra audio sources management

anabonne:
This is mostly about keeping track of all the extra audio sources across all trials and making sure that they are stopped in on_finish.

It involved a bit of refactor to put in common what could easily be factorized.


task-launcher/src/tasks/hearts-and-flowers/trials/instructions.js
```diff
@@ -128,6 +128,6 @@ function buildInstructionTrial(mascotImage, promptAudio, promptText, buttonText,
         <p>${buttonText}</p>
       </button>`.trim(),],
   };
-  overrideAudioTrialForReplayableAudio(trial, jsPsych.pluginAPI, replayButtonHtmlId);
+  overrideAudioTrialForReplayableAudio(trial, jsPsych.pluginAPI, replayButtonDivId);
   return trial;
 }
```
asengupta3, line 133: NIT: Add an extra line at EOF.

task-launcher/src/tasks/shared/trials/afcStimulus.js
```diff
@@ -23,31 +23,55 @@ let currPracticeChoiceMix = [];
...
+  extraAudioSources[extraAudioSourceKey] = audioSource;
+  audioSource.onended = () => {
+    if (extraAudioSources[extraAudioSourceKey]) {
+      extraAudioSources[extraAudioSourceKey] = null;
+    }
+  };
 };
 
-const showStaggeredBtnAndPlaySound = (btn) => {
+function showStaggeredBtnAndPlaySound(btn, jsPsychPluginApi, extraAudioSources) {
```
asengupta3, line 62: I personally prefer constants instead of function declarations because in JS, functions are hoisted which can lead to bad practices of calling functions before it is defined etc.

task-launcher/src/tasks/shared/helpers/replayAudio.js
```diff
@@ -1,39 +1,78 @@
...
-            const jsPsychAudioCtx = jsPsych.pluginAPI.audioContext();
+      const audioAsset = typeof trial.audioReplayOverrides.promptAudio === 'function'
+        ? trial.audioReplayOverrides.promptAudio()
+        : trial.audioReplayOverrides.promptAudio;
+      console.info(`replaying audioId=${audioAsset}`);
+      const audioBuffer = await trial.audioReplayOverrides.jsPsychPluginApi.getAudioBuffer(audioAsset);
 
-            isAudioPlaying = true;
+      const audioSource = jsPsychAudioCtx.createBufferSource();
+      audioSource.buffer = audioBuffer;
```
asengupta3, line 43: I think from a memory perspective this should be fine because your audioBuffers are being cached, meaning your memory should not go up more than the sum of all the audio buffers.

PR #95, opened 2024-05-10
Common UI components

anabonne:
Changes affecting a bunch of tasks.

- Added some common mixins, scss variables and styles to be used across our tasks.
- Lightly refactored some of our tasks to use these new common "stuff"
- Adjusted some dimensions in HaF task
- Adjusted  square buttons dimension and added cta/button style and animations for the following tasks:
  - Trog
  - egma-math

TODO as follow-ups:

- remove the arbitrary 80% width constraint on our "jspsych-content". This would allow more flexibility on our layouts and avoid issues like the line of 4 buttons suddenly not fitting and being off-center (which present PR fixes for above mentioned tasks for the moment). 
Note that HaF is ready for it and should not be impacted but this may not be the case of other tasks

- continue flattening the scss as much as possible by creating more ready-to-use, task&layout-agnostic, mixins and classes at the top level (avoid nesting as much as possible). Striving to limit their respective responsibilities as much as possible (e.g.: either a selector acts on font attributes or on container dimension, but not both).
Then you can either assign these common classes directly in javascript or, in the event you need a uniquely different styling, you can create a new top level scss class whose responsibility is solely to include common mixins and extend common class, and then to add just the stuff that is specific to its use-case.
- remove any fixed dimension in (px, rem, ...) and favor viewport units to prepare for smaller phone screens. (there can be exception on some small dimension where css does not behave properly like box-shadow.


task-launcher/src/tasks/hearts-and-flowers/helpers/utils.js
```diff
@@ -101,4 +101,8 @@ export const getStimulusLayout = (imageSrc, isLeft, promptText=undefined, replay
   `
 
   return template;
-}; 
+};
+
+export const inputButtonClasses = `response-btn cta-style-green`
```
anabonne, line 106: TODO follow-up: as mentioned in this PR #95's description, we want to get rid of these nested scss classes, here `response-btn` brings nothing unique except the dimension --> for which we should use a new top level scss class specific to HaF or re-use the `.square-button-dimensions` one

kachergis: @asengupta3 this is another one I'd value your input on: how should we organize the CSS?

asengupta3: @kachergis ideally what I would do is create a common file that holds all the common styles, like basic button styles, container styles, color palletes etc. And then, have separate files for each task where the styles are only applicable to those tasks. I think the way to go about it is to create the files and slowly start filling them out. We can also do a review of all tasks to make sure that we can at least move a few things to the common file. I can work with you next week on this.

anabonne: My two cents: As mentioned in the PR description. One or multiple scss files does not matter much. Since we are forced to compute the html and layout for each trial in JS anyway, I would refactor the scss to only have highly reusable components, styles, and font attributes while pushing all the task-specific stuff (like layout dimensions and how to fit 4 buttons in a row versus a diamond) to Javascript. You may find that you want to put the task-specific stuff into scss style but you may not...

anabonne: Another way to look at it is to split your styling along 3 concerns:

1. styling that is about "layout": arranging divs and other nodes between themselves and making sure things are shown where you expect them and fit on the screen. Ideally this is done in a responsive way.
2. styling that is about applying a theme, a color pallette, or an animation.
3. styling that defines common reusable component: e.g. as seen in this branch/PR, the cta buttons that appear everywhere and are shown on the bottom right corner of the figma next to the color palette. Or a square button.

I'd say separating 1 from 2 is much more important than #3 (which is more of an optimization). But one thing that I think is more important here is to remove all the nesting going on in the scss styles and in the big afcStimulus class -> this is what makes the code so brittle (a change for one screen breaks the styling in another one) if you can simply have separate declaration for each trial where each declare their own html layouts (via the prompt/stimulus and button_html jsPsych trial variables) then you'll be in a much better place: Then you can try to factorize common components and flesh out what is #3 above.

anabonne:
Just noticed commit [53793f1](https://github.com/levante-framework/core-tasks/pull/95/commits/53793f1b1cd5dd04bbe25792a26bf2ab6ffe90af) above and I think this is not supposed to be in this branch...

task-launcher/src/tasks/hearts-and-flowers/helpers/utils.js
asengupta3: There is a lot to unpack in the stuff that you mentioned but I think some of the things you mentioned are exactly what I was thinking but, we need a more organized path forward. My suggestion was to split them out so that they are easier to deal with in a sequential manner. Ideally the flow of styling should be:
1. Basic typography, color palettes, grid spacings, 
2. These should be then used to build buttons, inputs, containers and other layout element
3. You can then go on to use them however you want (per task vs overall styles)

However, at this stage we have a huge tech debt and too much inter-dependencies. We also do not have a standard style defined (I will work with our designer on that). So in the short term we need to unpack the dependencies first before we can make it more linear.

My greatest concern about making huge changes in a project like this is to create a bunch of regressions and spend a lot of time fixing it.

The path I am suggesting will have lot of small PRs and/ or new file additions (without breaking the current system) as opposed to a few huge PRs with a lot of code changes.

anabonne:
> Just noticed commit [53793f1](https://github.com/levante-framework/core-tasks/pull/95/commits/53793f1b1cd5dd04bbe25792a26bf2ab6ffe90af) above and I think this is not supposed to be in this branch...

Indeed. I just reverted 53793f1.
Initially I erroneously committed these changes to the common-ui-components branch under this commit 53793f1, which I then simply forgot to remove from the branch after I moved it to the extra-audio-playbacks branch.

task-launcher/src/tasks/hearts-and-flowers/helpers/utils.js
anabonne: Just to clarify: 
The regressions are already occurring: this is what the the Notion ticket [Fix-Linear-display-offset-match-UI](https://www.notion.so/Fix-Linear-display-offset-match-UI-2744735a25a843c29af309dd112fb3bd?pvs=4) is about: someone changes something for one afcStimulus trial and suddenly other trials covered by afcStimulus see their layout not fit on the screen anymore.
I'm not saying you need to make all the refactor all at once, instead I'd refactor one "afcStimulus" task at a time: extract its JS code in a separate class, refactor its css layout and styles to use new classes[^1], then do a QA pass on it and then move on to your next priority. After that this particular task will not regress when you change something in the nested scss code or afcStimulus.js code).

[^1]: Now when it comes to how much work you put in fleshing out these new css component and classes: this is where I would de-prioritize: we already have some figma design with some common components (font-size, buttons, color palettes, etc.) and I believe we can just make minimal tweaks to our existing css and layouts to get them to a point where all our screens layout and styles are based on composition rather than inheritance (i.e. no more nesting). Then you can always have more discussion on whether you want to refactor everything from flex to grid, or on getting a designer to do another pass on the figma. For now I'd focus on the small structural refactors.

PR #96, opened 2024-05-11
HaF: delay exiting the nested practice timeline to its feedback trial

anabonne:
When hitting the "winning-streak" on practice blocks (as of now it is 3 for mixed stimuli practice and 2 for the non-mixed)
We end the practice block nested timeline early via callbacks from the stimulus trial.
Problem: we skip the last feedback trial.

This makes sure this last feedback trial is shown before we exit the nested timeline early.


asengupta3 (commented):
This change looks good @kachergis. Looks like he is adding a callback on the `on_finish` to end the timeline explicitly.

PR #105, opened 2024-05-17
Removing clearAll call which breaks dashboard

Zio-4:


Zio-4:
This method wipes out the whole sessionStorage, not just stores created with this library. On the dashboard we keep some Pinia stores in sessionStorage which is why it breaks.

PR #115, opened 2024-05-29
Timer fix

Zio-4:
- Refactoring the timer to end the task when the timer is up, rather than waiting for a user response.
- Setting timer default to 100. Want to avoid setting null values for config. 
- Removing conditional code that is no longer needed
- Other minor adjustments around the timer.
- Added default values to parameter documentation

### DIFF
--- a/task-launcher/README.md
+++ b/task-launcher/README.md
@@ -66,25 +66,25 @@ In standalone web app mode, tasks and parameters can be changes through query st
 
 ### Common Parameters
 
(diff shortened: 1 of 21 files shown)


Zio-4:
@asengupta3 visibility tag. Removed a condition that's no longer needed that you recently implemented in skipping blocks in math.

kachergis (approved):
all looks good to me!

asengupta3:
@Zio-4 yes looks good to me. The logic is more centralized now. Thanks

PR #117, opened 2024-05-31
Save trial index property and all required fields with trial data

kachergis:
addresses a few tickets:
https://www.notion.so/Save-trialIndex-property-and-all-required-fields-with-trial-data-140ecedcade646f28793eb09f543da3c?pvs=4
https://www.notion.so/same-different-selection-expected_answer-is-No-answer-provided-and-is_correct-is-TRUE-9420dca0f20f483697d4f740422e8a64?pvs=4
https://www.notion.so/corpusTrialType-is-missing-for-number-line-trials-9a3ccd14e5b44d989fc38dc63160eace?pvs=4

### DIFF
--- a/task-launcher/src/tasks/math/trials/sliderStimulus.js
+++ b/task-launcher/src/tasks/math/trials/sliderStimulus.js
@@ -238,7 +238,6 @@ export const slider = {
       responseType: responseType,
       distractors: stimulus.distractors,
       corpusTrialType: stim.trialType,
-      trialIndex: stim.trialIndex,
       // slider_start: stimulus.item[1] === 1 ? sliderStart / 100 : sliderStart,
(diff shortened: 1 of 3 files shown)


task-launcher/src/tasks/math/trials/sliderStimulus.js
```diff
@@ -237,6 +237,8 @@ export const slider = {
       response: _toNumber(response),
       responseType: responseType,
       distractors: stimulus.distractors,
+      corpusTrialType: stim.trialType,
+      trialIndex: stim.trialIndex,
```
Zio-4, line 241: This is automatically saved, we don't need to list it here

Zio-4 (changes requested):

PR #123, opened 2024-06-12
SCSS 7-1 folder introduction

asengupta3:
This is just the introduction of 7-1 folder structure. No real code is affected. Current behaviors are not changed. The next PR will alter the css of the hearts and flowers task and will require more extensive testing.


Zio-4 (approved):
This is awesome! I'm excited to learn more about what SCSS can do. One small note, I noticed the overflow property is set to scroll. We don't want any scrolling, vertical or horizontal. Perhaps I'm not understanding the context though so please do correct me if i'm reading it wrong.

asengupta3:
@Zio-4 it was kept as a bail out, imagine you really shrink the screen and you can no longer place 4 buttons side by side. In that case, it would still work with scroll but for now, I have removed it but can be restored later.

PR #125, opened 2024-06-13
Hearts and Flowers styling refactor

asengupta3:
This uses the new styling convention to refactor the hearts and flowers task.

### DIFF
--- a/task-launcher/src/styles/abstracts/_index.scss
+++ b/task-launcher/src/styles/abstracts/_index.scss
@@ -2,3 +2,4 @@
 @forward 'variables';
 @forward 'fonts';
 @forward 'mixins';
+@forward 'spacings';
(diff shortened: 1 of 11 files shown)


Zio-4:
Here are a couple things I noticed on iPad air and iPad mini. Some of these may be fixed with new assets, right? @kachergis Should we fix this now or backlog it? Not sure what kind of devices partners are using. 

## iPad Mini:
Overflow-ipad-mini
- Header text intersects with replay button

audio-button-overlap-ipad-mini
- Header text intersects with replay button

## iPad Air:
button-ipad-air
- Continue button touches edge of screen.

asengupta3:
I guess I can reduce the paddings in a bunch of these things. But the way we have laid things out, I am sure in some device we are going to overflow (like a mobile phone in landscape mode). I think we should introduce scroll in such cases.

kachergis:
thanks, both! looks great, but would be good to reduce the padding between image and text, and button and text (and inside instruction box) -- really don't need much.

asengupta3:
Yeah its a good call because I think I have putting off standard spacing declarations. I will introduce that and look at the vertical layout a bit more. I was focused on the horizontal layout so much that I forgot about the vertical placement of things. Will work through it.

PR #132, opened 2024-06-20
[LHT-188] Add introductory instructions task

kachergis:


Zio-4 (approved):
LGTM. I expect we'll get feedback that it's blocks of text but we can change that in the future with some gifs.

PR #139, opened 2024-07-02
[LHT-371] update memory game instructions

zwatson2001:
Includes updates to the memory game: 

- text and audio prompts for each trial
- replay button for audio prompts during trials
- audio for feedback screen during practice trials 
- 1s delay between display and input phases of each trial


kachergis:
great work -- nearly there!
a couple minor things:
- replay button doesn't work for the "Touch the squares in the same order." prompt
- let's just combine the two sentences: "This time the game won't tell you if you're right or wrong. Let's start playing the memory game!" (ToDo for @kachergis )
- we need a little bit of padding between the stimulus and the response button on instruction trials
Screenshot 2024-07-05 at 2 24 11 PM

Screenshot 2024-07-05 at 2 22 51 PM

kachergis (changes requested):
I see the replay button, but it doesn't work on some trials (e.g. "Touch the squares in the same order")

kachergis:
I'm going to merge since it's already a lot better than before and I need to merge another PR, but @zwatson2001 could you wrap up the replay button fix on Mon -- thank you!

PR #141, opened 2024-07-05
Adding a way to disable the replay button for the initial stimulus time

asengupta3:
This PR addresses the following:
1. Disables the replay button when the initial stimulus audio is playing
2. Disables the replay button when the replay button is played and audio is playing
3. Centralizes the audio logic into a singleton so that audios don't overlap
4. Update the replay button styles to show that the button is disabled

### DIFF
--- a/task-launcher/src/styles/components/_buttons.scss
+++ b/task-launcher/src/styles/components/_buttons.scss
@@ -148,6 +148,17 @@ button {
       background-color: rgba($bg-button-primary-clicked, 0.6);
       border: 2px solid $border-button-primary-clicked;
     }
+
+    &:disabled {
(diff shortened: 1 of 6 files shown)


kachergis (changes requested):
this is 1000% better -- thanks so much, Aritra! I tested on all the multiAFC tasks. The one issue I've noticed is for practice trials (e.g. egma-math), if I click replay and then click a button, the feedback ("Please try again") will overlap with the prompt replay. (Ideally, it would stop the replay and just play the feedback.) I'm going to go ahead and merge this PR, and we can handle practice trials and the non multiAFC tasks early next week.

Zio-4 (approved):

PR #149, opened 2024-07-17
Consolidate practice button click logic

asengupta3:
1. Now keyboard and mouse click effectively call the same function
2. Should work with 2 choices and 4 choices

I have only tested it with egma-math with keyhelpers so far. But it might be worth testing other tasks.


kachergis (approved):
looks good to me! tried &keyHelpers=true in matrix-reasoning and trog

PR #154, opened 2024-07-29
Matrix reasoning button audio

zwatson2001:
Added the neutral 'select' audio to the response buttons and removed the audio prompts after the practice trials.

### DIFF
--- a/task-launcher/src/tasks/shared/trials/afcStimulus.js
+++ b/task-launcher/src/tasks/shared/trials/afcStimulus.js
@@ -46,7 +46,7 @@ function getStimulus(trialType) {
   if (!mediaAssets.audio[camelize(stim.audioFile)]) return mediaAssets.audio.nullAudio;
   // all tasks should have the replay button play whatever is in stim.audioFile (might be just prompt/instructions)
 
-  if (stim.task === 'Mental Rotation' && stim.notes !== 'practice' && stim.trialType !== 'instructions') {
+  if ((stim.task === 'Mental Rotation' || stim.task === 'Matrix Reasoning') && stim.notes !== 'practice' && stim.trialType !== 'instructions') {
     return mediaAssets.audio.nullAudio;
   }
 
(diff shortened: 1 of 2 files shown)


task-launcher/src/tasks/shared/trials/afcStimulus.js
```diff
@@ -46,7 +46,7 @@ function getStimulus(trialType) {
   if (!mediaAssets.audio[camelize(stim.audioFile)]) return mediaAssets.audio.nullAudio;
   // all tasks should have the replay button play whatever is in stim.audioFile (might be just prompt/instructions)
 
-  if (stim.task === 'Mental Rotation' && stim.notes !== 'practice' && stim.trialType !== 'instructions') {
+  if ((stim.task === 'Mental Rotation' || stim.task === 'Matrix Reasoning') && stim.notes !== 'practice' && stim.trialType !== 'instructions') {
     return mediaAssets.audio.nullAudio;
```
asengupta3, line 50: I think we should remove the nullAudio logic at some point. @Zio-4 this was done for some default work right? @zwatson2001 why do you want to return nullAudio in the case of matrix reasoning?

zwatson2001: That was to remove the audio prompt for matrix reasoning once the instructions and practice trials are over, but I could look into doing that a different way if that would be better.

kachergis: yes, for tasks such as matrix reasoning and mental rotation it is annoying to hear the same prompt repeated over and over for many trials. I'm fine with the nullAudio solution, but @asengupta3 if you prefer another approach please let us know.

asengupta3: I think we should get away from `jsPsychAudioMultiResponse` and change it to `jsPsychHTMLMultiResponse` and let the audio play in the on_load function. That will give us more control over overlapping audio etc. But that can be a separate PR. For now, let's merge this.

Zio-4 (approved):

PR #156, opened 2024-08-01
Staggered button reimplementation

asengupta3:
Feature updates:
1. Improved the layout animation from changing to visibility tag and not display (since display causes layout shift)
2. Made the buttons disabled during the staggered animation so that input can only be provided once all the options are presented.
3. Now we will start to pass specific layout configs from tasks to afcStimulus as opposed to afcStimulus making those decisions. This is an example of how we will do it.


Zio-4:
Getting error for each button when they render
Screenshot 2024-08-01 at 10 30 14 AM

Zio-4 (changes requested):
Above

asengupta3:
> Getting error for each button when they render Screenshot 2024-08-01 at 10 30 14 AM

Yeah this is expected @Zio-4 until we have the audio files for the emojis. This will not be merged till we have it.

Zio-4:
Going to convert it to Draft just so it's clear

PR #157, opened 2024-08-01
Update same different selection

zwatson2001:
I added feedback to the first trial of the second and third phases of the task ('something's the same' and the matching section). At the moment it uses the shared feedback trial, which displays a separate screen for feedback.


zwatson2001:
To do as of 8/2: Still need to make it so that feedback shows up at the beginning of the 2-match phase, and currently the feedback after the test dimensions phase does not repeat the trial if it is incorrect.

kachergis:
looking great! we may want to add the neutral sound when they make responses, and add a slight delay before starting the next trial.

PR #158, opened 2024-08-01
recrop HA and ToM stimuli to 9:7 (W:H) aspect ratio

kachergis:


asengupta3:
@kachergis what if we add whitespace on top and bottom of the images to make them 1:1 aspect ratio? Will that cause issues?

kachergis:
illustrator updated images, and we are keeping square aspect ratio (for now)

PR #164, opened 2024-08-06
Graceful handling of missing assets during staggered assets

asengupta3:
This implementation:
1. Does not stop the animation if either image assets or audio assets are missing for the staggered animation
2. It will just play null audio and move on with the animation

### DIFF
--- a/task-launcher/src/tasks/shared/trials/afcStimulus.js
+++ b/task-launcher/src/tasks/shared/trials/afcStimulus.js
@@ -28,26 +28,7 @@ let startTime;
 let keyboardFeedbackHandler;
 const incorrectPracticeResponses = [];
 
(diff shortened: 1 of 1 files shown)


task-launcher/src/tasks/shared/trials/afcStimulus.js
```diff
@@ -58,19 +39,22 @@ const showStaggeredBtnAndPlaySound2 = (
     'lev-staggered-opacity',
   );
   const img = btn.getElementsByTagName('img')?.[0];
```
asengupta3, line 41: @kachergis to test this locally replace this line by:
```
const img = btn.getElementsByTagName('imgNotFound')?.[0];
```

kachergis (approved):
looks great, thanks so much!

PR #168, opened 2024-08-09
Parameterizing the feedback component

asengupta3:
1. Some small button style update
2. Parameters on feedback component
3. Audio replay ends when trial ends for intro task to prevent overlap


task-launcher/src/tasks/same-different-selection/trials/stimulus.js
kachergis, line 1: I like using the larger size, but the images on `trialType='something-same-1'` trials need to be the same size as those on `something-same-2` trials:
Screenshot 2024-08-09 at 5 33 41 PM
Screenshot 2024-08-09 at 5 33 50 PM

kachergis (changes requested):
This is all great except that the cards on `something-same-1` trials are smaller than `something-same-2` trials, and need to match. They should probably also be `class='primary'`, but they have a class (`base-image-container`) that is currently used to disable them as buttons... @asengupta3 can you please find a solution?

PR #169, opened 2024-08-09
Remove text prompts from vocab and TROG tasks

zwatson2001:


task-launcher/src/tasks/trog/timeline.js
```diff
@@ -22,6 +22,7 @@ export default function buildTROGTimeline(config, mediaAssets) {
     responseAllowed: true,
     promptAboveButtons: true,
     task: config.task,
+    showPrompt: false
```
asengupta3, line 25: Can you make this a part of the layoutConfig parameter that you see above? But this is a great approach. We should not tie configs to visual logic.

task-launcher/src/tasks/shared/trials/afcStimulus.js
```diff
@@ -536,6 +537,7 @@ export const afcStimulus = ({ trialType, responseAllowed, promptAboveButtons, ta
         promptAboveButtons: promptAboveButtons,
         task: task,
         layoutConfig,
```
asengupta3, line 539: layoutConfig is an object which has layout info, so you can pass `showPrompt` as a key to that object.

task-launcher/src/tasks/trog/timeline.js
zwatson2001: Sure! I think I did this correctly in my last commit, but let me know if this wasn't what you had in mind and I can change it

kachergis (commented):
Looks good! Although I have an inkling of a memory that `promptAboveButtons` was intended to parameterize the same behavior... But that seems to no longer be implemented, and I think layoutConfig is a more reusable and thus preferable solution.

But we should probably make sure `promptAboveButtons` isn't being used anywhere, and if not, remove it.

task-launcher/src/tasks/shared/trials/afcStimulus.js
```diff
@@ -495,7 +496,7 @@ function doOnFinish(data, task) {
 }
 
 // { trialType, responseAllowed, promptAboveButtons, task }
-export const afcStimulusTemplate = ({ trialType, responseAllowed, promptAboveButtons, task, layoutConfig } = {}) => {
+export const afcStimulusTemplate = ({ trialType, responseAllowed, promptAboveButtons, task, layoutConfig, showPrompt } = {}) => {
   // TODO: pull out task-specific parameters (e.g., getPrompt(.., showPrompt=false) for Number Identification, TROG, ..)
```
kachergis, line 500: shouldn't `showPrompt` be embedded in `layoutConfig`?

kachergis:
@zwatson2001 merged, but leaving branch for you to look at my comments (and possibly update) on Mon -- thanks!

PR #170, opened 2024-08-09
Remove audio prompt after 3rd trial

zwatson2001:
Audio prompt should be there only for instructions and practice trials (with feedback) as well as the first three normal trials


task-launcher/src/tasks/shared/trials/afcStimulus.js
```diff
@@ -330,6 +336,10 @@ function doOnLoad(task, layoutConfig) {
   }
   const twoTrialsAgoStimulus = jsPsych.data.get().filter({ trial_index: twoTrialsAgoIndex }).values();
   
+  if (stim.trialType === 'Matrix Reasoning'){
+    matrixReasoningTrials += 1;
```
asengupta3, line 340: Do you think we can move this count increment to the timeline of Matrix reasoning itself and then pass a param from that in the layoutConfig object to not play the audio? I am trying to avoid branching in this function based on tasks and trial stages. They should be passed from their individual timelines.

PR #174, opened 2024-08-13
Making the vocab and TROG buttons larger

zwatson2001:
I created a new image-medium class that is larger than the standard one (hopefully without causing the buttons to go out of screen).

### DIFF
--- a/task-launcher/src/styles/abstracts/_variables.scss
+++ b/task-launcher/src/styles/abstracts/_variables.scss
@@ -14,4 +14,4 @@ $button-secondary-inner-shadow: 0px 0px 12px 0px;
 $response-size-default: clamp(120px, 12vw, 150px);
 $response-size-sm: clamp(100px, 20vw, 150px);
 $response-size-s: 25vw;
-$response-size-large: clamp(70px, 20vw, 250px)
+$response-size-m: clamp(70px, 20vw, 250px)
--- a/task-launcher/src/styles/abstracts/_variables.scss
+++ b/task-launcher/src/styles/abstracts/_variables.scss
@@ -14,4 +14,4 @@ $button-secondary-inner-shadow: 0px 0px 12px 0px;
 $response-size-default: clamp(120px, 12vw, 150px);
 $response-size-sm: clamp(100px, 20vw, 150px);
 $response-size-s: 25vw;
-$response-size-m: clamp(70px, 20vw, 250px)
+$response-size-m: calc(clamp(70px, 22vw, 250px) - 10px);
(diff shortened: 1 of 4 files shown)


task-launcher/src/styles/abstracts/_variables.scss
```diff
@@ -14,3 +14,4 @@ $button-secondary-inner-shadow: 0px 0px 12px 0px;
 $response-size-default: clamp(120px, 12vw, 150px);
 $response-size-sm: clamp(100px, 20vw, 150px);
 $response-size-s: 25vw;
+$response-size-large: clamp(70px, 20vw, 250px)
```
asengupta3, line 17: This looks still the same as `$response-size-sm`, only clamped at different pixels. We should probably increase the `20vw` values to justify the name also since it defined `image-medium`, it should probably be named `$response-size-m`.

task-launcher/src/styles/components/_buttons.scss
```diff
@@ -228,4 +228,30 @@ button {
...
+      border-radius: $lev-spacing-s;
+      float: left;
+    }
+
+    &:hover {
+      border-color: $bg-accent_secondary;
+    }
+
+    &:active {
+      border-color: $border-button-primary-clicked;
```
asengupta3, line 253: This will only change the default size and not affect the other screens. This should probably be tested across a range of screen sizes

task-launcher/src/tasks/shared/trials/afcStimulus.js
```diff
@@ -242,6 +242,8 @@ function getButtonHtml(task) {
   // TODO: add trial_type column to math item bank
   if (stimulus.trialType === 'instructions') {
     return "<button class='primary'>%choice%</button>";
+  } else if (task === 'vocab'|| task === 'trog'){
+    return "<button class='image-medium'>%choice%</button>"
```
asengupta3, line 246: We assign the button class in line `385`, we should probably keep all of it there so that when I update the layoutConfig, I can just do all of it in one place.

task-launcher/src/styles/abstracts/_variables.scss
zwatson2001: I think that value has to be at `20vw` or lower in order to prevent the buttons from going out of screen as far as I can tell, but I changed the name to `$response-size-m` (I agree that makes more sense)

task-launcher/src/styles/components/_buttons.scss
zwatson2001: I tested it with as many screen sizes as I could in Chrome and none of them cause the buttons to go out of screen, but let me know if it would be better to do it a different way!

asengupta3: @zwatson2001 pushed some changes to make the system a little more responsive. Please let me know if you have questions. I'll merge this tomorrow AM your time.

PR #175, opened 2024-08-14
Adding layoutConfig types

asengupta3:


asengupta3:
Just adding a types file so should not need a review. It was added before but was reverted back.

PR #176, opened 2024-08-15
Standardize end-of-task message

zwatson2001:
The goal of this PR is to display the same message at the end of each task (inside a prompt box with audio) regardless of whether it ended early due to mistakes. At the moment the screen displayed after making it all the way through a task has an exit button, but the other screen doesn't.

### DIFF
--- a/task-launcher/src/tasks/shared/trials/finishExperiment.js
+++ b/task-launcher/src/tasks/shared/trials/finishExperiment.js
@@ -17,7 +17,7 @@ export function finishExperiment() {
             window.removeEventListener('keydown', removeDOMElements);
         }
     }
-    window.addEventListener('click', removeDOMElements);S
+    window.addEventListener('click', removeDOMElements);
     window.addEventListener('keydown', removeDOMElements);
 
     jsPsych.endExperiment(
--- a/task-launcher/src/tasks/shared/trials/finishExperiment.js
+++ b/task-launcher/src/tasks/shared/trials/finishExperiment.js
@@ -1,12 +1,16 @@
 import { jsPsych } from "../../taskSetup";
-import { taskStore } from "../helpers";
+import { taskStore, PageAudioHandler } from "../helpers";
+import { mediaAssets} from '../../..'; 
 
 
 export function finishExperiment() {
     const t = taskStore().translations;
     jsPsych.endExperiment(
-      `<div id="prompt-container-text">
-                <p id="prompt">${t?.taskCompletion || 'Thank you for completing the task! You can now close the window.'}</p>
-            </div>`,
-    ); // ToDo: style text and add audio message?
+        `<div class='instructions-container'>
+            <div class='lev-row-container instruction'>
+                <h1 class='instructions-title'>${t.taskFinished}</h1>
+            </div>
+        </div>`,
+        PageAudioHandler.playAudio(mediaAssets.audio.taskFinished)
+    ); 
 }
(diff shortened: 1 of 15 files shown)


zwatson2001:
I updated the timelines of each task so that they skip all trials once max incorrect is reached (so that the last trial with the end message + exit button is always shown at the end).

asengupta3:
@zwatson2001 is there a reason that you are skipping trials with conditionals as opposed to finishing the whole experiment?

task-launcher/src/tasks/shared/trials/finishExperiment.js
```diff
@@ -1,12 +1,29 @@
...
+    ); 
+
+    window.addEventListener('click', (event) => {
+        const buttonId = event.target.id;
+        if (buttonId === 'exit-button') {
+            document.body.innerHTML = '';
+        }
+    }); 
+
+    document.addEventListener('keydown', () => {
```
asengupta3, line 26: This might add a permanent keydown event to destroy the entire html every time a key is placed. Why is this needed?

task-launcher/src/tasks/shared/trials/finishExperiment.js
```diff
@@ -1,12 +1,29 @@
...
+            </div>
+            <footer>${t.generalFooter}</footer>
+            <button id="exit-button" class="primary" style=margin-top:5%>Exit</button>
+        </div>`,
+        PageAudioHandler.playAudio(mediaAssets.audio.taskFinished)
+    ); 
+
+    window.addEventListener('click', (event) => {
+        const buttonId = event.target.id;
+        if (buttonId === 'exit-button') {
```
asengupta3, line 21: I forgot to mention that we should probably also remove the listener. So my suggestion would be to move this listener to before `jsPsych.endExperiment` is declared and remove the listener once the button is placed.
```
const removeDOMElements = (event) => {
        const buttonId = event.target.id;
        if (buttonId === 'exit-button') {
           document.body.innerHTML = '';
            // clean up so that it does not persist
           window.removeEventListener('click', removeDOMElements);
        }
}
window.addEventListener('click', removeDOMElements);
```

task-launcher/src/tasks/shared/trials/finishExperiment.js
```diff
@@ -1,12 +1,29 @@
...
 
 
 export function finishExperiment() {
     const t = taskStore().translations;
     jsPsych.endExperiment(
-      `<div id="prompt-container-text">
-                <p id="prompt">${t?.taskCompletion || 'Thank you for completing the task! You can now close the window.'}</p>
-            </div>`,
-    ); // ToDo: style text and add audio message?
+        `<div class='instructions-container'>
```
asengupta3, line 9: Can you replace this with `lev-stimulus-container`? We want to get rid of all `task.scss` declarations.

task-launcher/src/tasks/shared/trials/taskFinished.js
```diff
@@ -13,13 +14,17 @@ export const taskFinished = {
     const t = taskStore().translations;
     return `
         <div class='instructions-container'>
+          <div class='lev-row-container instruction'>
             <h1 class='instructions-title'>${t.taskFinished}</h1>
```
asengupta3, line 18: I think the class declaration can be removed.

task-launcher/src/tasks/shared/trials/finishExperiment.js
zwatson2001: Sure! I forgot to check for that on this file, thanks for the reminder

kachergis:
Hi both, thanks for all your work on this! What's the status here?

asengupta3:
@kachergis it is very close. There is just one more suggestion, I have. After that this will be ready to merge.

asengupta3 (approved):
LGTM! Thanks Zach for working on this and simplifying the logic. Please feel free to merge whenever you think you are done testing.

PR #185, opened 2024-08-29
layout config updates

asengupta3:


task-launcher/webpack.config.cjs
```diff
@@ -27,6 +27,7 @@ const commonConfig = {
   resolve: {
     fallback: {
       path: require.resolve('path-browserify'),
+      crypto: false,
```
Zio-4, line 30: Why do we need this?

Zio-4 (approved):

task-launcher/webpack.config.cjs
asengupta3: It was giving me an issue with webpack while running this locally. Interestingly , it was only happening intermittently but once it started it was consistent. I will keep this for now but I will keep testing it further and if I find a better solution, I will go with that.

PR #188, opened 2024-09-05
Layout config 3

asengupta3:


Zio-4 (changes requested):
Looks good but Intro is throwing an error when you complete it (ending screen). Could you also check if this is just for intro and not every task erroring at the ending screen?

Zio-4 (approved):

PR #189, opened 2024-09-10
Cypress tests for vocab and TROG

zwatson2001:
Adds Cypress tests for vocab and TROG that click the leftmost image button until the game ends.


kachergis:
Hi @zwatson2001 , looks great! I see a successful run via GH actions, but how do I run it locally? I tried `npx cypress run` and got:

```
Oops...we found an error preparing this test file:
  cypress/support/e2e.js

The error was:
  Error: Webpack Compilation Error
```

I'd like to merge this and have tests run automatically on commit, but would also like to know how to run tests locally. Thank you!

kachergis (approved):
Looks great! Just would like to know how to run tests locally.

zwatson2001:
Thanks @kachergis! Sorry I didn't see your comment until now. npx cypress run should be the right command - I wonder if the cypress-real-events package is the problem (that's the only thing I added to support/e2e.js). If you've already installed that though I'm not sure what's going on. I would be happy to look into it more if it still isn't working! 

I also just added a test for mental-rotation and updated all tests so that they select correct answers. Let me know if there's anything I should change before merging!

PR #192, opened 2024-09-13
Replay button should be present even after audio prompt stops playing 

zwatson2001:
This PR keeps the replay button on the screen even for trials where the audio prompt is silenced. It also prevents it from being disabled at first using the noAudio parameter stored in layoutConfigMap.


task-launcher/src/tasks/shared/helpers/replayAudio.ts
```diff
@@ -2,15 +2,17 @@ import { PageAudioHandler } from "./audioHandler";
 import { PageStateHandler } from "./PageStateHandler";
 
 
-export async function setupReplayAudio(pageStateHandler: PageStateHandler) {
+export async function setupReplayAudio(pageStateHandler: PageStateHandler, noAudio: boolean = false) {
```
asengupta3, line 5: You have the right approach. I am thinking whether we move this behavior to the `PageStateHandler` itself. Because the `PageStateHandler` is responsible for pagestate and `setupReplayAudio` function just listens to pagestate and decides the audio. Potentially we can do something like this:
In PageStateHandler.tsx:

```
constructor(audioFile: string, playStimulusOnLoad: boolean) {
    this.audioFile = audioFile;
    this.audioUri = mediaAssets.audio[camelize(this.audioFile)] ||
    mediaAssets.audio.nullAudio;
    this.getbuffer();
    this.replayBtn = document.getElementById('replay-btn-revisited') as HTMLButtonElement;
    this.playStimulusOnLoad = playStimulusOnLoad !== undefined ? playStimulusOnLoad : true;
  }
```

Initialize the PageStateHandler with the `noAudio` value like you did for `setupReplayAudio`

and then in this function:
```
if (pageStateHandler.replayBtn) {
    if (pageStateHandler.playStimulusOnLoad){
      ... rest of the code
    }
```

Let me know what you think.

asengupta3:
Also @kachergis will we ever have a trial where we do not show the replay button at all (like an instruction trial or otherwise)?

kachergis:
@asengupta3 there are no current tasks / instructions where we don't want the replay button, although I could imagine a language task where we don't want to allow participants to repeat the stimulus endlessly. (e.g., I don't think ROAR's Phoneme Awareness task allows repetition). Certainly default is to have the replay button.

task-launcher/src/tasks/shared/helpers/replayAudio.ts
kachergis: @zwatson2001 does Aritra's comment make sense? I'd like this one merged soon -- thanks!

asengupta3:
> @asengupta3 there are no current tasks / instructions where we don't want the replay button, although I could imagine a language task where we don't want to allow participants to repeat the stimulus endlessly. (e.g., I don't think ROAR's Phoneme Awareness task allows repetition). Certainly default is to have the replay button.

Okay then maybe we can rename our `noAudio` config to `playAudioOnLoad` and use it to determine whether we play the audio at the beginning or not.

zwatson2001:
@asengupta3 the playAudioOnLoad suggestion makes sense to me - @kachergis should I maybe wait to add that as a second PR so that this one can be merged quickly?

task-launcher/src/tasks/shared/trials/afcStimulus.ts
```diff
@@ -323,7 +323,8 @@ function doOnLoad(layoutConfigMap: Record<string, LayoutConfigType>) {
 
   const stim = taskStore().nextStimulus;
   const itemLayoutConfig = layoutConfigMap?.[stim.itemId];
-  const pageStateHandler = new PageStateHandler(stim.audioFile);
+  const noAudio = layoutConfigMap?.[stim.itemId].noAudio;
```
asengupta3, line 326: You can just do `itemLayoutConfig?.noAudio` here.

task-launcher/src/tasks/shared/trials/afcStimulus.ts
```diff
@@ -122,13 +123,12 @@ const getPromptTemplate = (
...
 ) => {
   let template = '<div class="lev-stimulus-container">';
-  if (!noAudio) {
-    template += `
-      <button id="${replayButtonHtmlId}" class="replay">
-        ${replayButtonSvg}
-      </button>
-    `;
-  }
+
```
asengupta3, line 126: I think this function no longer uses the `noAudio` so that can be removed.

asengupta3:
> @asengupta3 the playAudioOnLoad suggestion makes sense to me - @kachergis should I maybe wait to add that as a second PR so that this one can be merged quickly?

I think that is fine. The rename can wait till next PR.

task-launcher/src/tasks/shared/trials/afcStimulus.ts
zwatson2001: I forgot about that, thanks!

asengupta3 (approved):
LGTM!

PR #195, opened 2024-09-18
Rename noAudio to playAudioOnLoad

zwatson2001:
Renames the noAudio parameter in LayoutConfig to playAudioOnLoad


asengupta3 (approved):
LGTM! Thanks Zach!

PR #199, opened 2024-10-04
Update mental rotation mental rotation instructions

zwatson2001:
Adds the replay button to each instructions screen. I also used the mentalRotationInstruct4 audio in the corpus for the first practice trial.


kachergis:
looks good, thank you Zach!

PR #202, opened 2024-10-10
Intro task updates

zwatson2001:
I resized some of the elements in the instructions to prevent them from going off the screen - as far as I can tell using the device emulator in chrome, it now fits inside all ipad screens (as well as my own laptop). I also changed the message shown at the end of the task so that it makes more sense (it now says "instructions" rather than "game").

### DIFF
--- a/task-launcher/src/tasks/intro/trials/instructions.ts
+++ b/task-launcher/src/tasks/intro/trials/instructions.ts
@@ -38,13 +38,14 @@ export const instructions = instructionData.map(data => {
                     <button id="replay-btn-revisited" class="replay">
                         ${replayButtonSvg}
                     </button>
-                    <div class="lev-row-container instruction">
-                        <p>${t[data.prompt]}</p>
+                    <div class="lev-row-container instruction" style="width: 90vw;">
+                        <p style="font-size: 1.5em;">${t[data.prompt]}</p>
                     </div>
                     <div class="lev-stim-content-x-3">
                         <img
                             src=${mediaAssets.images[data.image]}
                             alt='Instruction graphic'
+                            style="height:330px;width:330px"
                         />
                     </div>
                 </div>
--- a/task-launcher/src/tasks/shared/trials/taskFinished.js
+++ b/task-launcher/src/tasks/shared/trials/taskFinished.js
@@ -2,33 +2,39 @@ import jsPsychHTMLMultiResponse from '@jspsych-contrib/plugin-html-multi-respons
 import { taskStore, PageAudioHandler } from '../helpers';
 import { mediaAssets } from '../../..';
 
-export const taskFinished = {
-  type: jsPsychHTMLMultiResponse,
-  data: () => {
-    return {
-      // save_trial: true,
-      assessment_stage: 'instructions',
-    };
-  },
-  stimulus: () => {
-    const t = taskStore().translations;
-    return `
-        <div class='lev-stimulus-container'>
-          <div class='lev-row-container instruction'>
-            <h1>${t.taskFinished}</h1>
-          </div>
-            <footer>${t.generalFooter}</footer>
-        </div>`;
-  },
-  button_choices: [`Continue`],
-  keyboard_choices: 'ALL_KEYS',
-  button_html: '<button class="primary" style=margin-top:10%>Exit</button>',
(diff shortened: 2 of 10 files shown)


task-launcher/src/tasks/shared/trials/taskFinished.js
```diff
@@ -2,33 +2,39 @@ import jsPsychHTMLMultiResponse from '@jspsych-contrib/plugin-html-multi-respons
...
+              <h1>${t[endMessage]}</h1>
+            </div>
+              <footer>${t.generalFooter}</footer>
+          </div>`;
+    },
+    button_choices: [`Continue`],
+    keyboard_choices: 'ALL_KEYS',
+    button_html: '<button class="primary" style=margin-top:10%>Exit</button>',
+    on_load: () => {
+      const endMessage = intro ? 'introFinished' : 'taskFinished';
```
asengupta3, line 30: Can we rename the mediaAsset to `taskFinished` instead of `introFinished`?

task-launcher/src/tasks/intro/trials/instructions.ts
```diff
@@ -38,13 +38,14 @@ export const instructions = instructionData.map(data => {
                     <button id="replay-btn-revisited" class="replay">
                         ${replayButtonSvg}
                     </button>
-                    <div class="lev-row-container instruction">
-                        <p>${t[data.prompt]}</p>
+                    <div class="lev-row-container instruction" style="width: 90vw;">
+                        <p style="font-size: 1.5em;">${t[data.prompt]}</p>
```
asengupta3, line 42: I would prefer if these went into the class definitions. You can add a new definition of font size called:
`font-size-xs-em: 1.5em` and then update the:
```
&.instruction {
    display: flex;
    padding: $lev-spacing-s;
    justify-content: center;
    align-items: center;
    text-align: center;
    border: 3px solid $bg-accent_secondary;
    border-radius: 32px;
    background-color: $bg-bubble-base;
   width: 90vw;

    p {
      font-size: $font-size-xs-em;
      font-weight: $font-weight-light;
      line-height: 1.0;
      margin: 0;
    }
```
Does the width need to be 90vw? or should the max-width be 90vw? We want consistent experience so we will need to just make sure it looks good across different tasks.

task-launcher/src/tasks/shared/trials/taskFinished.js
zwatson2001: `taskFinished` is the asset that all the other tasks use - the text is "You've completed the game. Thank you!" and `introFinished` is a separate asset for intro only that I just created (it says "You've finished the instructions. Thank you!"). I could definitely rename `introFinished`, but I think it has to be different from `taskFinished`. Sorry if I'm misunderstanding your comment though!

task-launcher/src/tasks/intro/trials/instructions.ts
zwatson2001: That makes sense, I'll work on adding it to the class definition. I think setting max-width to 90vw would probably be better, so I'll try that - thanks!

task-launcher/src/tasks/shared/trials/taskFinished.js
asengupta3: Makes sense. I guess when we pull them both in the common folder that we have been talking about, both can exist. So a config param makes sense. I am wondering though for future proofing should we pass the asset name as the parameter to `taskFinished` as opposed to supporting just two use cases. If we pass the asset name then in the future we can support as many ending screens as we want. But that is optional, I'll leave it up to you.

task-launcher/src/tasks/intro/trials/instructions.ts
zwatson2001: I ended up making an `instructions-small` class with the font size and max-width changes and using that for intro - I thought making the font smaller for all tasks didn't look as good. Let me know if you would prefer a different design though!

task-launcher/src/tasks/shared/trials/taskFinished.js
zwatson2001: That makes sense, I switched it so that we now pass in the asset name.

asengupta3 (approved):
LGTM

PR #206, opened 2024-10-15
Fix some response times not being saved in data

zwatson2001:
Response times were not being saved for some trials - as far as I can tell from looking at the jsPsych data object, these were the number line trials with buttons and the matching and test-dimensions phases of SDS. In this PR I manually calculated response times for these trials and added them to jsPsych.data (I saw this was also being done for keyboard responses in afcStimulus). I also noticed that RTs are not saved for practice trials - I'm not sure if we want those, but I could save them if that would be good!

### DIFF
--- a/task-launcher/src/tasks/same-different-selection/trials/stimulus.ts
+++ b/task-launcher/src/tasks/same-different-selection/trials/stimulus.ts
@@ -203,13 +203,15 @@ export const stimulus = {
         response: choices[data.button_response],
       });
 
-      if (stim.trialType === "test-dimensions") {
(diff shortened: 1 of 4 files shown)


task-launcher/src/tasks/same-different-selection/trials/stimulus.ts
```diff
@@ -200,6 +203,14 @@ export const stimulus = {
         response: choices[data.button_response],
       });
 
+      if (stim.trialType === "test-dimensions") {
```
asengupta3, line 206: NIT: `'test-dimensions'`, single quotes.

PR #207, opened 2024-10-17
Update CSS to prevent content from leaving screen

zwatson2001:
The goal of this PR is to prevent the bottom buttons from leaving the screen, which was happening for matrix reasoning and TOM. I applied the `instruction-small` class to these tasks and also made each of the size variables for the `image` button class slightly smaller, which changes the size of the big stimulus images since they were previously set to 3x the size of the buttons. As far as I know `image` is currently only used for the buttons in these two tasks, so buttons in other tasks shouldn't be affected. I also slightly decreased the font size used for `instruction-small` and gave it min and max values.

### DIFF
--- a/task-launcher/src/styles/abstracts/_fonts.scss
+++ b/task-launcher/src/styles/abstracts/_fonts.scss
@@ -4,9 +4,9 @@ $font-size-l: 64px;
 $font-size-m: 48px;
 $font-size-s: 40px;
 $font-size-xs: 32px;
-$font-size-xs-em: 1.5em; 
 $font-size-xxs: 24px;
 $font-size-xxxs: 16px;
+$font-size-xs-em: clamp($font-size-xxxs, 1.4em, $font-size-xs); 
 
 $font-weight-dark: 700;
 $font-weight-bold: 600;
--- a/task-launcher/src/styles/abstracts/_variables.scss
+++ b/task-launcher/src/styles/abstracts/_variables.scss
@@ -11,7 +11,7 @@ $button-default-font-weight: 500;
 $button-secondary-box-shadow: -8px -10px 0px 0px;
 $button-secondary-inner-shadow: 0px 0px 12px 0px;
 
-$response-size-default: clamp(120px, 12vw, 150px);
-$response-size-sm: clamp(100px, 20vw, 150px);
-$response-size-s: 25vw;
+$response-size-default: clamp(120px, 10vw, 150px);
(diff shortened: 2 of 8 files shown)


asengupta3:
@zwatson2001 I am personally fine with these changes but I fear that it may not be solving the underlying issue of height here. Even though it might be looking fine for your device, if someone else has a different aspect ratio screen, the buttons might go out of screen again. One way to avoid this is by introducing scroll. We can either scroll the instruction container or the entire layout. But before we do that, it might be helpful to discuss with Anastasia to get an idea of what might be aesthetically better.
This is a very tricky part of web development with no winning solution. There is always going to be some compromise here but lets discuss those compromises and make it clear and document what works and what does not work.

zwatson2001:
@asengupta3 that all sounds good, thanks for taking a look! I'd be happy to discuss what looks best, and I can leave this PR open until we make a decision.

kachergis:
I'm pretty firm on not introducing scrolling -- not feasible for young kids. I know there's no silver bullet, but we need a solution that scales element widths based on screenwidth, and element height based on screenheight. If this is better than current, we should merge it...

mcfrank:
I agree with George about scrolling being off the table, so scaling is the
right solution... I don't know enough about the current approach to give
good suggestions, but it seems reasonable to me to think about scaling down
the buttons for small devices.

asengupta3:
All of that is possible except the instruction prompt size. You can always break your scaled systems by introducing a very long prompt. Regardless of how much we scale the elements, the instruction element height is always an unknown and I believe Zach is trying to fix that in this PR. Some of the buttons are probably going out of screen because of very large instructions. I think if we do not want scrolling, one way to handle that is to introduce some sort of word count limit to the instructions. That way we will know what is the maximum amount of scaling required for the fonts. The rest of the elements can be auto scaled to the screen.

zwatson2001:
There was already some auto scaling of buttons in place (I think there are three sizes corresponding to small, medium, and large screens). This PR just makes the button smallest button size and the instructions a bit smaller. I think if we're looking for a more permanent solution then introducing a word limit per instruction trial makes sense, because very long instructions would probably be easier to understand if they were broken up with different supporting images anyway. I think I'll test and merge this for now, and continue to look out for adjustments to scaling that should be made as the tasks develop. Thank you all for your input!

mcfrank:
I like the word limit idea, Zach. That seems like it gives us a guarantee
that we stay on the page. :)

PR #209, opened 2024-10-21
Record replay presses

zwatson2001:
This PR saves the number of times the replay button was pressed as `audioButtonPresses` for each trial.


asengupta3:
I would put this in the `audioHandler.ts` as a static member variable instead of a global variable in setupReplayAudio. That way in the future we can track how many times each audio were played if there are multiple audio on the page etc.

zwatson2001:
@asengupta3 that makes sense, thanks! Is this what you had in mind?

task-launcher/src/tasks/shared/helpers/replayAudio.ts
```diff
@@ -19,8 +20,9 @@ export async function setupReplayAudio(pageStateHandler: PageStateHandler) {
     }
 
     async function replayAudio() {
+      PageAudioHandler.replayPresses ++;
       pageStateHandler.disableReplayBtn();
-      PageAudioHandler.playAudio(pageStateHandler.audioUri, onAudioEnd);
+      PageAudioHandler.playAudio(pageStateHandler.audioUri, onAudioEnd);
```
asengupta3, line 25: NIT: Extra space.

task-launcher/src/tasks/shared/helpers/audioHandler.ts
```diff
@@ -9,6 +9,13 @@ export class PageAudioHandler {
   static audioContext: BaseAudioContext;
   static audioUri: string;
   static audioSource?: AudioBufferSourceNode;
+  static replayPresses: number;
+
+  static saveReplayPresses(){
+    jsPsych.data.addDataToLastTrial({
+      audioButtonPresses: PageAudioHandler.replayPresses
```
asengupta3, line 16: Hmm, so we are making the `PageAudioHandler` aware to the data saving to the backend? I am fine with it, usually I like to keep these classes stateless. In this case that would mean instead of calling `saveReplayButton()`, I would just do: `jsPsych.data.addDataToLastTrial({ audioButtonPresses: PageAudioHandler.replayPresses})` .

But I won't be very fussy about this :). I will leave it up to you. I am fine either way.

asengupta3 (approved):
LGTM!

task-launcher/src/tasks/shared/helpers/audioHandler.ts
zwatson2001: That sounds good, thanks! I was going back and forth about which way would make more sense.

PR #211, opened 2024-10-22
log response location for all tasks

zwatson2001:
Saves response location as an index (0-3 if there are 4 buttons) for all tasks.


asengupta3 (approved):
LGTM!

PR #217, opened 2024-11-05
Fix audio issue caused by audioMultiResponse plugin

zwatson2001:
This PR switches to using the html-multi-response plugin in afcStimulus to avoid issues caused by the audio-multi-response plugin. The audio is now played using PageAudioHandler once the page loads.

### DIFF
--- a/task-launcher/src/tasks/shared/trials/afcStimulus.ts
+++ b/task-launcher/src/tasks/shared/trials/afcStimulus.ts
@@ -75,7 +75,7 @@ const showStaggeredBtnAndPlaySound = (
 
 const handleStaggeredButtons = async (layoutConfig: LayoutConfigType, pageState: PageStateHandler) => {
   if (layoutConfig?.isStaggered) {
-      const parentResponseDiv = document.getElementById('jspsych-audio-multi-response-btngroup') as HTMLDivElement;
+      const parentResponseDiv = document.getElementById('jspsych-html-multi-response-btngroup') as HTMLDivElement;
       let i = 0;
       const stimulusDuration = await pageState.getStimulusDurationMs();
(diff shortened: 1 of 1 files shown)


task-launcher/src/tasks/shared/trials/afcStimulus.ts
```diff
@@ -1,5 +1,5 @@
 // For all tasks except: H&F, Memory Game, Same Different Selection
-import jsPsychAudioMultiResponse from '@jspsych-contrib/plugin-audio-multi-response';
+import jsPsychHtmlMultiResponse from '@jspsych-contrib/plugin-html-multi-response';
```
asengupta3, line 2: This looks good. I think we can just replace `@jspsych-contrib/plugin-audio-multi-response` everywhere including SDS and other non afc tasks. I assume the issue will eventually happen in other tasks as well. Before we merge we should deploy this to qa and test it.

zwatson2001:
I was not able to use `jspsych-html-multi-response` in the SDS `afcMatch.ts` file because for some reason ending the trial manually after the second card is selected causes an error in the html plugin code (but the single-answer trials work fine with this plugin). I left the audio plugin there for now because it isn't currently causing any problems, but @asengupta3 if you have time and interest feel free to look into how the html plugin might be used for the afc match trials. Thanks!

asengupta3:
@zwatson2001 does this fix the audio not auto playing in Safari or is this just primarily removing the audio plugin.

zwatson2001:
@asengupta3 this is just removing the audio plugin, although I think it also prevents audio overlapping between trials in some places - I haven't made much progress on the safari autoplay issue yet unfortunately. I think I'll merge this once you've merged your taskStore() PR if that sounds good.

PR #224, opened 2024-11-25
Improvements to Inference

Emily-ejag:


task-launcher/serve/serve.js
kachergis, line 1: Can you define the `numItemsPerStory` parameter? The meaning isn't transparent to me from the name. In general we'd like to ad generic parameters (that are useful in multiple tasks, not specific to only one task). e.g., if the meaning is something like number of items per block, then I'd rather have itemsPerBlock. (By this same reasoning, I'd like to change/remove `inferenceNumStories` -- why not use e.g. `stimulusBlocks`?)

task-launcher/src/tasks/shared/helpers/config.ts
kachergis, line 1: Again, would rather reuse an existing parameter or add a reusable one (e.g. `itemsPerBlock`) rather than add one-off parameters (that also don't refer to the specific task they are used by).

task-launcher/src/tasks/shared/helpers/taskStore.js
kachergis, line 1: Renaming `corpusId` to `corpus` would be a breaking change for many tasks -- these names are defined in the column headings of the item banks.

task-launcher/src/tasks/shared/trials/setup.js
kachergis, line 1: I'm not sure this would behave as we want it to: @zwatson2001 don't sometimes have `stimulus` defined on trials that we don't want to show anything (including a fixation)? e.g. in number identification trials?

kachergis (changes requested):
We can't rename `corpusId` to `corpus`, and I'm not sure if there would be side effects of the default fixation trial setup, but otherwise this looks fine. Our team is on vacation through Sunday, so we won't have any time for testing until Monday morning.

task-launcher/src/tasks/shared/trials/setup.js
zwatson2001: As far as I know stimulus is never defined for the setup/fixation trials (unless I'm forgetting something). It's been on my list to remove the fixation cross from some tasks, and this actually might be a good way of doing that that we could use. But I'll test it thoroughly on Monday to make sure it isn't breaking any of the tasks!

asengupta3:
Additionally and in relation to @kachergis comment about reusing variables, we had introduced two variables in the previous PR:
```
story: (() => {
        if (row.task === 'roar-inference') {
          return row.story;
        } else {
          return '';
        }
      })(),
      storyId: (() => {
        if (row.task === 'roar-inference') {
          return row.story_id;
        } else {
          return '';
        }
      })(),
```
Since the previous PR was a rush we merged this change but we should reuse our variables and not create new one off ones. I suggest, we use `item` and `itemId` instead of creating `story` and `storyId`. Ideally your corpus should have those columns defined. The `item` should have a key value which is then read from the translation file to get the actual story value.

task-launcher/src/tasks/shared/trials/setup.js
Emily-ejag: Thanks @zwatson2001 , Yes we wanted to remove the cross, or change it for other thing if necessary. Let me know if you need me to change something here

task-launcher/src/tasks/shared/helpers/taskStore.js
Emily-ejag: I reviewed that the only apps using this are trog and inference so far, and both were writing "undefined" in the database since there is no corpusId. I can change it back to corpusId and make some changes in inference and trog if necessary.

task-launcher/src/tasks/shared/helpers/config.ts
Emily-ejag: I reused it with a little change, so it can fit 1 itemPerStory as default for inference

task-launcher/src/tasks/shared/helpers/getCorpus.js
Emily-ejag, line 1: @asengupta3 I think I am using now the correct params, can you please double check?

task-launcher/src/tasks/shared/helpers/taskStore.js
Emily-ejag: Done!, but I had to add corpus as well if corpusId is not there

task-launcher/src/tasks/shared/helpers/config.ts
```diff
@@ -91,7 +91,7 @@ export const setSharedConfig = async (firekit: RoarAppkit, gameParams: GameParam
     buttonLayout: buttonLayout || 'default',
     numberOfTrials: numberOfTrials ?? 300,
     task: taskName ?? 'egma-math',
-    stimulusBlocks: stimulusBlocks ?? 3,
+    stimulusBlocks: stimulusBlocks ?? (taskName === 'roar-inference' ? 1 : 3),
```
asengupta3, line 94: I would just use the url param to control this value instead of changing the default for a specific task.

task-launcher/src/tasks/shared/helpers/taskStore.js
```diff
@@ -62,7 +62,7 @@ export const setTaskStore = (config) => {
     // -----
     audioFeedback: config.audioFeedback,
     skipInstructions: config.skipInstructions,
-    corpusId: config.corpusId,
+    corpusId: config.corpusId ?? config.corpus,
```
asengupta3, line 65: Where is the `corpus` value coming from? Is it from the actual corpus excel sheet? Can we simply rename the column to `corpusId` then?

Emily-ejag: It is not coming from a column on the excel sheet, it the the actual name of the corpus we are pulling. I tried different ways to get it, and I actually think we don't need `corpusId`, just `corpus`

task-launcher/src/tasks/trog/helpers/scores.js
```diff
@@ -51,12 +51,14 @@ export const computedScoreCallback = (rawScores) => {
...
   // But we also want to update the total score so we add up all of the others.
   //const totalScore = _reduce(_omit(computedScores, ['composite']), (sum, score) => sum + score.subScore, 0);
-  const { totalCorrect, trialNumTotal } = taskStore();
+  const { totalCorrect } = taskStore();
 
   computedScores.composite = {
     totalCorrect: totalCorrect,
-    totalNumAttempted: trialNumTotal,
-    totalPercentCorrect: Math.round((totalCorrect / trialNumTotal) * 100),
+    totalNumAttempted: rawScores?.composite?.test?.numAttempted,
```
asengupta3, line 58: Is this functionality only being used by ROAR?

task-launcher/serve/firebaseConfig.js
```diff
@@ -19,4 +19,4 @@ const productionFirebaseConfig = {
   appId: '1:46792247600:web:ea20e1fe94e0541dd5a0f5',
 };
 
-export const firebaseConfig = ENV === 'production' ? productionFirebaseConfig : devFirebaseConfig;
+export const firebaseConfig = ENV === 'production' ? productionFirebaseConfig : devFirebaseConfig;
```
asengupta3, line 22: NIT: EOF missing.

task-launcher/src/tasks/trog/helpers/scores.js
Emily-ejag: Yes, just for Roar apps. It was writing infinity to our database, which gave us an error on Redivis. The actual data is the one on raw

task-launcher/src/tasks/shared/helpers/getCorpus.js
```diff
@@ -75,20 +75,6 @@ const transformCSV = (csvInput, numOfPracticeTrials, sequentialStimulus, task) =
...
-        }
-      })(),
-      storyId: (() => {
-        if (row.task === 'roar-inference') {
-          return row.story_id;
-        } else {
-          return '';
-        }
-      })(),
     };
```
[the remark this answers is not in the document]
Emily-ejag: Yes!

task-launcher/serve/firebaseConfig.js
Emily-ejag: yeah, not sure what this is here, since no changes were made

task-launcher/src/tasks/shared/helpers/config.ts
asengupta3: Correct. Is that an issue? We add url params all the time, including taskName etc

task-launcher/src/tasks/shared/helpers/taskStore.js
asengupta3: Okay got it. In that case, please add a new variable called `corpus` and just assign the corpus to that instead of assigning it to the corpusId.

task-launcher/serve/firebaseConfig.js
asengupta3: You can revert to the one that was there in the main branch

task-launcher/src/tasks/shared/helpers/taskStore.js
Emily-ejag: Sure!

task-launcher/src/tasks/shared/helpers/config.ts
Emily-ejag: Alright!

task-launcher/serve/firebaseConfig.js
Emily-ejag: Done!

task-launcher/src/tasks/shared/helpers/getTranslations.js
zwatson2001, line 1: @kachergis and I just noticed this change - we'd like to keep this so that the tasks are always pulling from `https://storage.googleapis.com/road-dashboard/item-bank-translations.csv` (which we've been overwriting whenever there is a change) rather than adding separate dated versions of the translations sheet. I'll make a PR to change it back, but @Emily-ejag I think this means you will have to add to item-bank-translations.csv with any inference-related changes that you have in the newer versions from Nov 7th and 22nd. Sorry not to flag this earlier!

asengupta3 (approved):
@kachergis the changes looks good to me. There are some additions to the taskStore object but I think that is fine for now. We should revisit the `trialNumTotal` and the `stimulusBlocks` values. I think they can either be removed or be provided through another existing variable.

task-launcher/src/tasks/shared/helpers/getTranslations.js
```diff
@@ -51,7 +51,7 @@ export const getTranslations = async (configLanguage) => {
   async function fetchData() {
     const urls = [
       // This will eventually be split into separate files
-      `https://storage.googleapis.com/road-dashboard/item-bank-translations-2024-11-07.csv`,
+      `https://storage.googleapis.com/road-dashboard/item-bank-translations-2024-11-22.csv`,
```
zwatson2001, line 54:
```suggestion
      `https://storage.googleapis.com/road-dashboard/item-bank-translations`,
```

zwatson2001 (changes requested):

task-launcher/src/tasks/shared/helpers/getTranslations.js
Emily-ejag: Oh I think we discussed it with @kachergis so we always update the versions and we know when those were changed. I am happy to change it to `item-bank-translations.csv` and add the new stuff, but I think it is good in case of an emergency, to have separate versions?

zwatson2001: I think we'd be open to eventually having a versioning system, but for now we'd like to keep just one file since we're frequently having to make changes to it. We typically edit the item bank translations google doc and then upload that to the bucket. Thanks for switching everything over!

zwatson2001 (approved):

task-launcher/src/tasks/shared/helpers/getTranslations.js
kachergis: Hi Emily, IIRC we said it's good to rename (and thus save) the previous versions with the date they were created, but the newest one will always be item-bank-translations.csv (and any changes your side makes needs to be up-to-date in the Google sheet, since otherwise our side (e.g. Amy) will not know if a dated version is updated in core-tasks). I do think we will want to split out CSVs for different languages as we add more, and ultimately an in-code versioning system is probably also good -- but not currently part of our process. Thank you!

task-launcher/src/tasks/roar-inference/trials/afcInference.ts
kachergis, line 1: I still don't understand why you can't change `corpus` to `corpus_id`; the goal of the mono-repo is to standardize as much as possible... but of course this is specific to ROAR inference, so it won't interfere with other tasks -- just makes it harder to keep track of using different parameters for different tasks.

task-launcher/src/tasks/shared/helpers/getCorpus.js
kachergis: @asengupta3 and @zwatson2001 has this been tested?

task-launcher/src/tasks/roar-inference/trials/afcInference.ts
Emily-ejag: I tried many times, but it gives `undefined` for `corpusId`. Also, I really don't know where you use it, and I even suggested to not use `corpusId` at all. I tested in inference and in trog (those where actually using corpusId before and writing `undefined` to our database for a long time)

PR #225, opened 2024-11-25
Enable CAT for vocab, TROG, mental rotation, and matrix reasoning 

zwatson2001:
This PR enables CAT for vocab, TROG, mental rotation, and matrix reasoning. It can be turned on or off for each of these tasks with with url parameter `&cat=true/false`. The corpus must have a difficulty column for CAT to work correctly. The current timeline for a task running with CAT turned on is structured like this:

- One block including all instruction + practice trials 
- A CAT block that selects from all items with difficulty defined - starts with 5 randomly selected validated items
- A block with 5 randomly selected unvalidated items

A participant will skip from the CAT block to the unvalidated block after reaching the stopping rule, which is currently SEM < 0.3 (this is a placeholder value that can be changed). This currently replaces the usual stopping rule of 3 incorrect trials when CAT is running.


task-launcher/serve/serve.js
```diff
@@ -48,11 +48,12 @@ const pid = urlParams.get('pid');
 const inferenceNumStories = urlParams.get('inferenceNumStories') === null? null : parseInt(urlParams.get('inferenceNumStories'), 10);
 
 // Boolean parameters
-const keyHelpers = stringToBoolean(urlParams.get('keyHelpers')); // GK: shouldn't this default to false?
+const keyHelpers = stringToBoolean(urlParams.get('keyHelpers')); // GK: shouldn't this default to false
```
asengupta3, line 51: This does default to false, so I think we can remove the comment.

PR #228, opened 2024-12-04
Fix memory game UI

zwatson2001:
This PR addresses the comment from Germany that the OK button is leaving the screen during the instruction trials. I shrank the video size to create more room, which seems to solve the problem for most screen sizes. 

There was another comment about images briefly flashing on the screen before the instruction video. Because this stops happening after the task is reloaded a few times, I think it might be related to how the assets are loading (although I tried preloading the assets for memory game and it did not fix the issue). For now I added a short gap between instruction screens, which I think makes them easier to read and makes the image flickering less distracting when it does happen.

### DIFF
--- a/task-launcher/src/styles/layout/_video-containers.scss
+++ b/task-launcher/src/styles/layout/_video-containers.scss
@@ -6,6 +6,11 @@ $instruction_m_size: 50vw;
 $instruction_sm_size: 70vw;
 $instruction_s_size: 80vw;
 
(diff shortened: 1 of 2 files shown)


asengupta3 (approved):
@zwatson2001 changes look good to me, I would just make sure that @kachergis is fine with it. I think a lot of them were done to keep consistency across tasks but it looks like it is time to be bit a more flexible.

kachergis:
LGTM -- thanks, both!

PR #229, opened 2024-12-04
revert to original translation sheet

zwatson2001:


kachergis:
@Emily-ejag , note that we keep `item-bank-translations.csv` as the latest version of translations (for all tasks), so please make sure you add any ROAR task translation changes to the corresponding Google doc.  We're open to splitting out ROAR and LEVANTE tasks to use separate CSVs if that's better for you.

PR #231, opened 2024-12-06
add 3D instruction screen to mental rotation

zwatson2001:
Adds a text-only instruction screen signaling that 3D items will be shown. It will be placed before the first 3D item regardless of where it is in the timeline, so it should work even with CAT turned on. I'll wait to merge this until the translation has been reviewed and uploaded.


kachergis (approved):
looks good to me! I wonder if we should make instruction screens automatically skip if a translation is missing...it seems like it might be nice to be robust to some errors that are bound to happen (as we need to backfill more instuctions for the downward extension...)

mcfrank:
good idea about the skipping behavior!

PR #234, opened 2024-12-11
Limit items in random cat blocks

zwatson2001:
This PR excludes 3D trials from the starting block in the mental rotation cat. It also fixes a bug that involved blank cells in the difficulty corpus column defaulting to difficulty=0 in the tasks.


task-launcher/src/tasks/shared/trials/afcStimulus.ts
```diff
@@ -450,7 +450,7 @@ function doOnFinish(data: any, task: string, layoutConfigMap: Record<string, Lay
       target = response.target; 
       data.correct = responseValue === target;
     }
-
+
```
asengupta3, line 453: NIT: Remove this line

asengupta3 (approved):
LGTM!

PR #236, opened 2024-12-17
Prevent typescript errors in getTranslations.ts

zwatson2001:


asengupta3:
@zwatson2001 I think you will need to run `npm install` again. I installed this package `"@types/papaparse": "^5.3.15",` that introduces the type for `papaparse`.

zwatson2001:
Thanks @asengupta3! I'll just close this then - it seems like it isn't necessary

PR #238, opened 2024-12-18
Define pulse animation and disable buttons/key presses

zwatson2001:
This PR fixes a bug in hearts and flowers caused by the user clicking the buttons during the incorrect feedback screen (for some reason this was causing the task to skip additional trials while the screen was still showing). I disabled the buttons and all key presses during feedback trials to prevent this, and also added the pulse animation back into the H&F button class.


asengupta3 (approved):
LGTM! Thanks.

PR #240, opened 2024-12-20
adding tasks background fallback

Zio-4:
Adding background color (color matched from image) property as a fallback as [recommended](https://developer.mozilla.org/en-US/docs/Web/CSS/background-image) in MDN docs. Temporary fix for the white background of tasks in the dashboard (Something to do with PrimeVue v4's theming I suspect).


task-launcher/src/styles/base/_jspsych.scss
```diff
@@ -13,6 +13,7 @@
   -ms-user-select: none; /* Internet Explorer/Edge */
   user-select: none; /* Non-prefixed version, currently supported by Chrome, Edge, Opera and Firefox */
   color: #0f3cac;
+  background-color: #D7E3FD;
```
asengupta3, line 16: Can you use the variable `$bg-workspace-base`?

task-launcher/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@levante-framework/core-tasks",
   "version": "1.0.0-beta.28",
   "description": "",
-  "main": "lib/index.d3531f49.js",
-  "module": "lib/index.d3531f49.js",
+  "main": "lib/index.a06ba69a.js",
```
asengupta3, line 5: Why does need to change?

asengupta3 (approved):
LGTM!

PR #248, opened 2025-01-29
SDS downward extension changes

zwatson2001:
This PR adds a 'younger-kid mode' to same different selection, which is turned on with `heavyInstructions = true` in the url. This involves using a different set of practice and instruction trials (which have animations and different prompt wording). Younger kid instructions are currently identified in the corpus with `difficulty < 0`. I also added a screen at the end asking the instructor to assess the quality of the data from the task run. 

Once this is approved, I will modify the default SDS corpus to add in the new downward extension instructions/practice trials before I merge it. For now, you can see what the changes look like by running the task with `corpus=sds-downward-extension-test`.


task-launcher/src/tasks/same-different-selection/trials/heavyInstructions.ts
```diff
@@ -0,0 +1,260 @@
+import jsPsychHtmlMultiResponse from '@jspsych-contrib/plugin-html-multi-response';
+import { mediaAssets } from '../../..';
+//@ts-ignore
```
asengupta3, line 3: You should need ts-ignore anymore.

task-launcher/src/tasks/same-different-selection/trials/heavyInstructions.ts
```diff
@@ -0,0 +1,260 @@
...
+import { PageStateHandler, replayButtonSvg, setupReplayAudio, PageAudioHandler, camelize } from '../../shared/helpers';
+//@ts-ignore
```
asengupta3, line 5: You should not need ts-ignore anymore

task-launcher/src/tasks/shared/trials/afcStimulus.ts
```diff
@@ -12,6 +12,7 @@ import {
   camelize,
   setSentryContext,
 } from '../helpers';
+import { handleStaggeredButtons } from '../helpers/staggerButtons';
```
asengupta3, line 15: Add this to the index.ts of helpers and then you can import it directly from `../helpers`

task-launcher/src/tasks/shared/helpers/prepareCat.ts
```diff
@@ -57,9 +83,9 @@ export function selectNItems(corpus: StimulusType[], n: number) {
     const trial = shuffledTrials.pop();
 
     if (trial !== undefined) {
-      finalTrials.push(trial); 
+      finalTrials.push(trial);
     }
   }
 
-  return finalTrials; 
+  return finalTrials;
```
asengupta3, line 90: NIT: EOF

task-launcher/src/tasks/same-different-selection/timeline.ts
```diff
@@ -67,58 +85,135 @@ export default function buildSameDifferentTimeline(config: Record<string, any>,
...
+  if (heavy) {
+    timeline.push(somethingSameDemo1)
+    timeline.push(somethingSameDemo2)
+    timeline.push(somethingSameDemo3)
   }
 
-  // test-dimensions phase
-  for (let i = 0; i < phase2b; i++) { 
+  // push in remaining instructions
+  instructionPracticeBlock2.forEach(trial => {
```
asengupta3, line 156: Move this right below:
```
 instructionPractice.filter(trial => trial.blockIndex == 0).forEach(trial => {
    timeline.push(ipBlock(trial))
  });
```
and add a comment about adding ipblock and what is doing to the whole timeline. That will be easier to read.

task-launcher/src/tasks/same-different-selection/timeline.ts
```diff
@@ -67,58 +85,135 @@ export default function buildSameDifferentTimeline(config: Record<string, any>,
...
+  // something-same
+  for (let i = 0; i < block2; i++) {
     timeline.push(setupStimulus)
     timeline.push(stimulusBlock)
+    timeline.push(buttonNoise) // adds button noise for appropriate trials
+    timeline.push(dataQualityBlock)
+    //timeline.push(feedbackBlock)
+  }
+
+  instructionPractice.filter(trial => trial.blockIndex == 3).forEach(trial => {
```
asengupta3, line 169: Move this below:
```
instructionPractice.filter(trial => trial.blockIndex == 0).forEach(trial => {
    timeline.push(ipBlock(trial))
  });
```

task-launcher/src/tasks/same-different-selection/timeline.ts
```diff
@@ -67,58 +85,135 @@ export default function buildSameDifferentTimeline(config: Record<string, any>,
...
   const timeline = [
     preloadTrials, 
     initialTimeline, 
   ];
 
-  const { phase1, phase2a, phase2b, phase2c, phase2d, phase2e } = sdsPhaseCount
+  let block1 = 0, block2 = 0, block3 = 0, block4 = 0, block5 = 0, block6 = 0;
+
+  taskStore().corpora.stimulus.forEach((trial: StimulusType) => {
+    switch (trial.blockIndex) {
```
asengupta3, line 105: Instead of introducing a bunch of variables, you can use an array to store the count. So this who case logic can turn into:
```
const blockCountList = [];
blockCountList[trial.blockIndex] = (blockCountList[trial.blockIndex]  || 0) + 1;
```
Then you can create a bunch of functions for each of the block operations and put them in another array. So for example, for block 4 you have:
```
function testDimensionUpdate() {
    timeline.push(setupStimulus)
    timeline.push(stimulusBlock) 
}
```
Put this function inside the timeline manipulation so that the variables like `setupStimulus` is already defined. Now you create an operation array:
```
const blockOperations = [fn1, fn2, fn3, testDimensionUpdate, ...];
```

`fn1..fn3` are representative functions for other blocks. Now you can do:
```
blockCountList.forEach((count, index) => {
    for (let i = 0; i < count; i += 1) {
      blockOperations[index]();
    }
});
```
This will make the code look cleaner and more scalable when/ if we add new blocks. We will just need to add a block manipulation logic and push it to `blockOperations`.

asengupta3:
@zwatson2001 nice work. I have added some suggestions around how to organize the code a bit more. Please let me know if you have questions. I may have missed a few things, I can take a look once more after you have addressed the comments. There are few EOF missing and you can safely remove `ts-ignore` now. If that is causing any issues, please let me know.

task-launcher/src/tasks/shared/trials/afcStimulus.ts
zwatson2001: That's helpful, thanks!

zwatson2001:
@asengupta3 thanks for taking a look! Those suggestions all sound good to me - I'll let you know once I've implemented them.

PR #251, opened 2025-01-30
CAT updates for downward extension

zwatson2001:
This PR adds a url parameter to set the starting theta value, which can be used to start out younger kids with easier items for matrix reasoning and mental rotation. For right now, I also limited the difficulty of the start items so that they are less than the starting theta (otherwise hard items could still show up in the block of 5 random starting items, which I forgot about during the LEVANTE team meeting). 

Another approach could be to create the option to remove the starting items entirely. I'm also not sure if we always want to limit the starting item difficulty - for older kids we might want to allow items that are higher than the starting theta. @mcfrank let me know if you have any thoughts about what to do with the starting block for younger kids. Thanks!


zwatson2001:
We decided to always limit the starting block so that difficulty is <= 0 for now (regardless of what the starting theta is set to)

mcfrank:
I think that's a very good decision.

PR #252, opened 2025-02-04
Math CAT

zwatson2001:
This PR implements an adaptive version of math with 3 blocks. The CAT currently runs 50% of the trials in the corpus from each block before advancing to the next block, carrying over a single theta estimate between blocks. The first block starts with 5 random items restricted to only trials from that block.  Any practice and instruction trials in the corpus are added to the beginning of their respective block.

### DIFF
--- a/task-launcher/src/tasks/math/timeline.ts
+++ b/task-launcher/src/tasks/math/timeline.ts
@@ -114,7 +114,6 @@ export default function buildMathTimeline(config: Record<string, any>, mediaAsse
     const fullCorpus = prepareCorpus(corpus); 
 
     const allBlocks: StimulusType[][] = prepareMultiBlockCat(taskStore().corpora.stimulus); 
-    console.log(allBlocks);
 
     const newCorpora = {
       practice: taskStore().corpora.practice,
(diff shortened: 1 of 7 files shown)


task-launcher/src/tasks/math/timeline.ts
```diff
@@ -36,6 +38,9 @@ export default function buildMathTimeline(config: Record<string, any>, mediaAsse
   const translations: Record<string, string> = taskStore().translations;
   const validationErrorMap: Record<string, string> = {};
 
+  const { runCat } = taskStore(); 
+  const { semThreshold } = taskStore();
```
asengupta3, line 42: You can do:
`const { runCat, semThreshold } = taskStore();`

task-launcher/src/tasks/shared/helpers/index.ts
```diff
@@ -24,7 +24,7 @@ export * from './setSkipCurrentBlock'
 export { PageAudioHandler } from './audioHandler';
 export { PageStateHandler } from './PageStateHandler';
 export { camelize } from './camelize';
-export { prepareCorpus, selectNItems } from './prepareCat';
+export * from './prepareCat';
```
asengupta3, line 27: Let's do explicit exports so that we know exactly what is being exported from this file itself.

asengupta3:
I have not tested this code but it looks good for the most part in terms of the code

asengupta3 (approved):
LGTM!

PR #253, opened 2025-02-07
create adult reasoning task

zwatson2001:
This PR implements the adult reasoning task. Because it shares most of its assets with math, it currently uses the same google cloud bucket. I also added the `secondary--wide` button class to accommodate the longer response options.


asengupta3 (approved):
LGTM! Just make sure the files have EOF lines.

PR #254, opened 2025-02-11
Update downward extension animations for SDS

zwatson2001:
This PR adds a new button animation to the downward extension version of SDS.


asengupta3 (approved):
LGTM! Maybe once we finalize the animations, we can move some of these to css.

PR #256, opened 2025-02-18
Fix scoring of higher-dimension matching trials in same different selection

zwatson2001:
This PR fixes the way that higher-dimension matching trials are scored. Because the item names in the corpus leave out the number or color dimension if it has the default value (for example,  "med-red-square" rather than "med-red-square-1-white") I added those values back in the task code before they are scored. Additionally, a selection is now flagged as a repeat only if the same two cards were picked in one previous selection (rather than across multiple previous selections).

### DIFF
--- a/task-launcher/src/tasks/same-different-selection/trials/afcMatch.ts
+++ b/task-launcher/src/tasks/same-different-selection/trials/afcMatch.ts
@@ -220,8 +220,8 @@ export const afcMatch = {
           previousSelections.forEach((item: string[]) => {
             // check that most recent selection does not have the same cards as a previous selection (even in reverse)
             if (
-              selections[0] == item[0] && selections[1] == item[1] ||
-              selections[1] == item[0] && selections[0] == item[1]
+              selections[0] === item[0] && selections[1] === item[1] ||
+              selections[1] === item[0] && selections[0] === item[1]
             ) {
               hasNewSelection = false; 
             }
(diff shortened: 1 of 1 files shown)


task-launcher/src/tasks/same-different-selection/trials/afcMatch.ts
```diff
@@ -158,11 +158,24 @@ export const afcMatch = {
       previousSelections = [];
     }
 
+    function cleanAttributes(attributes: string[]) {
```
asengupta3, line 161: @kachergis there were other defaults right? Or was it just color?

asengupta3:
Great! This was a bug that I remember discussing with George. Thanks for taking a look at it and proposing a solution. IMO I think the fix is updating the corpus to not be ambiguous and assume the client knows the default values. Imagine a database having ambiguous values and expecting clients to assume defaults. over multiple clients (browser, iOS app, Android app). But if that is not possible, then your solution works.

task-launcher/src/tasks/same-different-selection/trials/afcMatch.ts
zwatson2001: I think the dimensions that have defaults are background color and number.

zwatson2001:
@asengupta3 thanks for taking a look - that makes sense to me! @kachergis do you think it would be okay to change the names of response options in the corpus so that they always specify all the dimensions? I can apply those changes and remove the corresponding part of this PR if so. Thanks!

task-launcher/src/tasks/same-different-selection/trials/afcMatch.ts
kachergis: yes, the dimensions with defaults are number (1) and background color (white) -- adding them back is a good solution!

kachergis:
I see Aritra's point, but I would rather fix this in code, as there are more manual steps (and thus chance of error) needed to replace the image files, update the corpus, and change the analysis pipeline.

PR #257, opened 2025-02-19
feat(): Update the practice trial behavior of memory game

asengupta3:
Currently if the final practice trial has wrong input, we show a try again page but continue on the task. In this PR we update that to not show that if the final trial is wrong for the second round of tests. Here is the summary of the new behavior:
1. For the first round of practice trials (3 total), we provide for each trial.
2. If the final trial (3rd trial) is wrong, we start another set of 3 practice trials
3. In this second round, if the final trial answer is correct we provide the correct feedback screen. If the final trial answer is wrong we do not show the feedback screen and go straight to the test trials.


zwatson2001 (approved):
This looks good to me! Thanks for implementing this

PR #261, opened 2025-02-25
SDS cypress test

zwatson2001:


task-launcher/src/tasks/same-different-selection/trials/stimulus.ts
```diff
@@ -172,9 +172,21 @@ export const stimulus = (trial?: StimulusType) => {
     const buttonContainer = document.getElementById('jspsych-html-multi-response-btngroup') as HTMLDivElement;
     buttonContainer.classList.add('lev-response-row');
     buttonContainer.classList.add('multi-4');
+
     const trialType = stimulus.trialType; 
     const assessmentStage = stimulus.assessmentStage;
 
+    // if the task is running in a cypress test, the correct answer should be indicated with 'correct' class
+    if (window.Cypress && trialType !== 'something-same-1') {
```
asengupta3, line 180: I wonder if there is a way to test things without changing the business logic only for tests. Let's discuss it this week. I think I may have some idea. Not sure if it will work though.

asengupta3 (approved):
Let's merge this, and I can separately look into making it more generic/ independent of DOM changes done only during tests.

PR #263, opened 2025-03-03
Use age to set de param

zwatson2001:
This PR calculates the user's age from the information in user params and sets `heavyInstructions` to true if the user's age is 5 or less.


task-launcher/src/tasks/shared/helpers/getAge.ts
```diff
@@ -0,0 +1,11 @@
...
+    const today = new Date(); 
+
+    const years = today.getFullYear() - birthYear; 
+    const months = (today.getMonth() + 1) - birthMonth; 
+
+    const ageInMonths = (12 * years) + months; 
+    const age = Math.floor(ageInMonths / 12); 
+
+    return age; 
+}
```
asengupta3, line 11: NIT: EOF

task-launcher/serve/serve.js
```diff
@@ -70,7 +70,7 @@ async function startWebApp() {
       };
 
       const userParams = {
-        pid,
+        pid
```
asengupta3, line 73: Leave the `,` since the lint might eventually complain.

task-launcher/src/tasks/shared/helpers/config.ts
```diff
@@ -61,6 +62,8 @@ export const setSharedConfig = async (firekit: RoarAppkit, gameParams: GameParam
 
   const {
     userMetadata = {},
+    birthMonth, 
+    birthYear,
```
asengupta3, line 66: So is the goal that we will send the `birthMonth` and `birthYear` as url param and perform the age calculation here?

zwatson2001: The dashboard code was already set up to create the TaskLauncher object with birth month and year stored from the user's initial registration as part of user params (this is the PR I made on the dashboard side: https://github.com/levante-framework/levante-dashboard/pull/76/commits/0d38403c5b3bc4bf651e32eed62c75023501d8f5), so I don't think we will need to specify those in the url.

asengupta3: Great!

PR #264, opened 2025-03-11
Add practice trials to math and shorten gap between blocks

zwatson2001:
This PR allows practice trials to be added to the slider block, which currently repeat once as a block if they are answered incorrectly. It also prevents the pause that occurs when the task skips to the next block.

### DIFF
--- a/task-launcher/src/tasks/math/timeline.ts
+++ b/task-launcher/src/tasks/math/timeline.ts
@@ -1,7 +1,14 @@
 import 'regenerator-runtime/runtime';
 import store from 'store2';
 // setup
-import { initTrialSaving, initTimeline, createPreloadTrials, prepareCorpus, prepareMultiBlockCat } from '../shared/helpers';
+import { 
+  initTrialSaving, 
+  initTimeline, 
+  createPreloadTrials, 
+  prepareCorpus, 
+  prepareMultiBlockCat, 
+  getStimulus 
+} from '../shared/helpers';
 import { jsPsych, initializeCat } from '../taskSetup';
 import { slider } from './trials/sliderStimulus';
 import { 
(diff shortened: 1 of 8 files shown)


task-launcher/src/tasks/math/timeline.ts
```diff
@@ -76,26 +84,118 @@ export default function buildMathTimeline(config: Record<string, any>, mediaAsse
...
 
   const sliderBlock = (trial?: StimulusType) => {
     return {
-      timeline: [slider(trial)],
+      timeline: [
+        slider(trial), 
+        feedbackBlock(trial)
+      ],
       conditional_function: () => {
+        const trialsSkipped = store.page.get("trialsSkipped");
```
asengupta3, line 153: Instead of using the `store.page` @Zio-4 had written a wrapper called `taskStore` to centralize all that logic. We should add it there and use all the operations using `taskStore()`

task-launcher/src/tasks/math/trials/sliderStimulus.ts
```diff
@@ -243,23 +257,46 @@ export const slider = (trial?: StimulusType) => {
     const stimulus = trial || taskStore().nextStimulus; 
     const responseType = stimulus.trialType.includes('4afc') ? 'button' : 'slider';
 
-    setUpAudio(responseType)
+    setUpAudio(responseType); 
+
+    if (stimulus.assessmentStage === "practice_response") {
+        const practiceBtns: NodeListOf<HTMLButtonElement> = document.querySelectorAll('.practice-btn');
```
asengupta3, line 263: I think we use a very similar logic in `afcStimulus`. Do you think we can centralize all of this in one place?

task-launcher/src/tasks/shared/trials/afcStimulus.ts
```diff
@@ -216,7 +216,18 @@ function handlePracticeButtonPress(
   isKeyBoardResponse: boolean,
   responsevalue: string | number,
 ) {
-  const choice = btn?.children?.length ? (btn.children[0] as HTMLImageElement).alt : btn.textContent;
+  let choice; 
+  if (stim.trialType.includes("Fraction")) {
+    // get the numbers nested inside the button in the DOM
+    const numerator = [0, 0, 0, 0, 0].reduce((el: any, index) => el.children[index], btn).textContent; 
+    const denominator = [0, 0, 0, 1, 0].reduce((el: any, index) => el.children[index], btn).textContent;
```
asengupta3, line 223: Additionally this logic is being reused below so I suggest, pulling this out in a function and just simply calling it from both places.

PR #269, opened 2025-03-25
prevent muting of repetitive audio when triggered by replay button

zwatson2001:


zwatson2001:
@asengupta3 sure, sorry not to document this at all. The purpose of this PR is to prevent repetitive audio from being muted if it is triggered by the replay button. The PageAudioHandler class mutes audio after it has replayed it twice, meaning that if the user clicks the replay button more than once it will be muted on subsequent replays. I think the user should be allowed to replay the audio as many times as they want, even if the prompt audio itself is muted when it is the same across multiple consecutive trials

PR #270, opened 2025-03-25
Adding new styles and fixing responsiveness

asengupta3:
This PR introduces:
1. Similar style to range input for Number Line 4afc vs Number Line Slider even if the slider is disabled (tested it some for Firefox as well but it is not as similar looking as Chrome and other webkit browser).
2. Responsive button view and support for diamond layout and proper placement of key helpers

Some screenshots from the new implementation:
Screenshot 2025-03-26 at 1 56 15 AM
Screenshot 2025-03-26 at 1 56 44 AM

Screenshot 2025-03-26 at 1 56 55 AM

### DIFF
--- a/task-launcher/src/styles/pages/_egma-math.scss
+++ b/task-launcher/src/styles/pages/_egma-math.scss
@@ -22,3 +22,48 @@
   display: flex;
   justify-content: space-around;
 }
+
+.number-line-4afc-slider {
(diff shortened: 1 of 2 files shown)


zwatson2001 (approved):
I tested it and it looks good to me!

Zio-4 (approved):

PR #271, opened 2025-03-25
Update math feedback and instructions

zwatson2001:
This PR allows instruction trials in the math corpus to include images, which is necessary for the slider block instruction trial. It also creates the option to not show the prompt at the bottom of the feedback screen.


asengupta3 (approved):
LGTM!

PR #272, opened 2025-03-26
Same different selection CAT

zwatson2001:
This PR allows same different selection to be run in a CAT. Because this task has blocks of trials that have to stay together sequentially (for example, `something-same-1` and `something-same-2` trials or `first-response`, `second-response`, `third-response` in the matching phase) only the first trial of each block is selected in a CAT. The following trials in the block are set at the end of the first trial.


task-launcher/src/tasks/shared/helpers/prepareCat.ts
```diff
@@ -113,16 +139,16 @@ export function updateTheta(item: StimulusType, correct: boolean) {
...
-      b: item.difficulty, // item difficulty (from corpus)
-      c: item.chanceLevel, // probability of correct answer from guessing
-      d: 1, // max probability of correct response (default 1)
-    };
-
-    if (!Number.isNaN(zeta.b) && item.assessmentStage !== 'practice_response') {
-      const answer = correct ? 1 : 0;
-      cat.updateAbilityEstimate(zeta, answer);
-    }
+      const zeta = {
```
asengupta3, line 142: There is extra tab here for some reason. Please remove that. You can also run `npx run prettier` to get some standard formatting.

task-launcher/src/tasks/same-different-selection/timeline.ts
```diff
@@ -98,8 +110,43 @@ export default function buildSameDifferentTimeline(config: Record<string, any>,
...
+  }) :
   taskStore().corpora.stimulus.forEach((trial: StimulusType) => {
-    blockCountList[Number(trial.blockIndex)] = (blockCountList[Number(trial.blockIndex)] || 0) + 1;
+    // if not running a CAT, trials are blocked by their type 
+    let trialBlock: number; 
+    switch (trial.trialType) {
+      case "test-dimensions":
+          trialBlock = testDimPhase2 ? 3 : 0; 
+        break;
+      case "something-same-1":
```
asengupta3, line 126: Let's move this to its own function, called `setTrialBlockValue`.

task-launcher/src/tasks/same-different-selection/timeline.ts
```diff
@@ -163,8 +210,59 @@ export default function buildSameDifferentTimeline(config: Record<string, any>,
     }
 
     // push in test trials
-    for (let i = 0; i < count; i += 1) {
-      blockOperations[index]();
+    if (cat) {
+      // returns timeline object containing the appropriate trials - only runs if they match what is in taskStore
+      function runCatTrials(trialNum: number, trialType: 'stimulus' | 'afc') {
```
asengupta3, line 215: Let's move this function out of here and put in the global scope.

task-launcher/src/tasks/shared/helpers/prepareCat.ts
```diff
@@ -93,8 +97,30 @@ export function selectNItems(corpus: StimulusType[], n: number) {
 // separates cat corpus into blocks
 export function prepareMultiBlockCat(corpus: StimulusType[]) {
   const blockList: StimulusType[][] = []; // a list of blocks, each containing trials
+  let newCorpus: StimulusType[] = [];
+  
+  if (taskStore().task === "same-different-selection") {
```
asengupta3, line 102: Let's not overload this function further. We can create a local function for SDS for prepareMultiBlockCat in the SDS folder. We are not sharing any behavior here so it does not make sense to use this function.

asengupta3 (changes requested):
@zwatson2001 looking at code bifurcation for `cat` related changes, do you think we just create a completely separate `timeline` and `stimulus` if cat is enabled? Right now each of these files are becoming big and will be hard to debug in the future. If we split them out we can tackle them separately. Let me know what you think.

digital-pro:
Not directly related, but the AI models also get really grumpy processing large files, as I learned yesterday😊

zwatson2001:
@asengupta3 I agree that splitting out timelines for cat and non-cat versions makes sense. I made this change for SDS - let me know if this is what you had in mind. I can also do this for the other CAT tasks in a separate PR if you think that would be good

task-launcher/src/tasks/taskConfig.ts
```diff
@@ -61,7 +63,10 @@ export default {
     setConfig: setSharedConfig,
     getCorpus: getCorpus,
     getTranslations: getTranslations,
-    buildTaskTimeline: sameDifferentSelectionTimeline,
+    buildTaskTimeline: (cat: boolean, config: Record<string, any>, mediaAssets: MediaAssetsType) => 
+      cat ?
```
asengupta3, line 67: Instead of passing `cat` you can do:
```
taskStore().runCat ? sameDifferentSelectionTimelineCat(config, mediaAssets) 
:  sameDifferentSelectionTimeline(config, mediaAssets),
```

This way you don't have to change the signature of buildTaskTimeline

PR #286, opened 2025-03-27
Inference Landscape Fontsize

Emily-ejag:
Hello, we have had some issues when using ipad on landscape for `roar-inference`. This small PR corrects the css on the `roar-inference-button`, `instructions`, and` questions`.

THEN: 
image
image

NOW: 
image
image

### DIFF
--- a/task-launcher/src/styles/pages/_roar-inference.scss
+++ b/task-launcher/src/styles/pages/_roar-inference.scss
@@ -1,0 +1,78 @@
+@use '../abstracts/' as *;
+
+.roar-inference-btn {
+    display: flex;
+    width: 40vh;
+    height: 9vh;
+    padding: 8px;
+    justify-content: center;
+    align-items: center;
+    gap: 8px;
+    border: 1px solid black;
+    background: white;
+    border-radius: 32px;
+    // text
+    color: black;
+    font-family: 'Noto Sans';
+    font-size: 1rem;
+    font-style: normal;
+    font-weight: 700;
+    line-height: normal;
+    cursor: pointer;
+
+    &:hover {
+      background-color: gainsboro;
+    }
+
+    @include respond-to('huge') {
+      font-size: 1.5rem;
+    }
+}
+
+.instruction-no-border{
+    display: flex;
+    border: none;
+    border-radius: 0;
+    background-color: transparent;
+    color: black;
+    font-size: x-large;
+    padding-bottom: 0;
+    margin-right: 25%;
+    margin-left: 25%;
+
+    p{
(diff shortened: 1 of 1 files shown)


digital-pro:
It makes sense to me to make the screen more compact. In that vein, we had our Columbia site using tablets with 1340x800 displays, and have suggested to them that landscape mode is the preferred orientation. Do we know if after this change the screens would work in that situation?

Emily-ejag:
> It makes sense to me to make the screen more compact. In that vein, we had our Columbia site using tablets with 1340x800 displays, and have suggested to them that landscape mode is the preferred orientation. Do we know if after this change the screens would work in that situation?

It will work for `roar-inference` for now

task-launcher/src/styles/components/_buttons.scss
```diff
@@ -65,14 +65,19 @@ button {
     // text
     color: black;
     font-family: 'Noto Sans';
-    font-size: 1.5rem;
```
asengupta3, line 68: Please don't update default styles. They will affect all tasks. You can create your own class and modify it however you want.

task-launcher/src/styles/layout/_containers.scss
```diff
@@ -117,6 +117,10 @@
       margin-left: 5%;
       margin-right: 5%;
     }
+
+    @include respond-to('huge') {
+      font-size: xx-large;
+    }
   }
 
   &.roar-instruction-question{
```
asengupta3, line 126: These are not generic classes and do not belong in generic definitions. Please move all your stylings to a file called `_roar-inference.scss` and place it under `styles -> pages`. You create your custom classes there and modify them however you want.

task-launcher/src/styles/components/_buttons.scss
Emily-ejag: Hi @asengupta3 , this is the `roar-inference-btn` class that was created only and applied only to the inference buttons

task-launcher/src/styles/layout/_containers.scss
Emily-ejag: same case here, the instruction-no-border if you remember is only created and applied to `roar-inference`

task-launcher/src/styles/components/_buttons.scss
Emily-ejag: this is not a default style

asengupta3 (approved):
Looks great! Thanks @Emily-ejag

zwatson2001 (approved):
Looks good to me!

PR #287, opened 2025-03-27
Require button press or keydown on end screen to finish task

zwatson2001:
The goal of this PR is to standardize the user's experience at the end of the task. Currently, the user is sometimes returned to the dashboard before they press the 'Exit' button on the end screen (particularly when they error/time out of a task), and sometimes only after they press it. This PR requires that the task does not end until the button is pressed, regardless of how the task is completed.

### DIFF
--- a/task-launcher/src/index.ts
+++ b/task-launcher/src/index.ts
@@ -9,6 +9,7 @@ import './styles/index.scss';
 import taskConfig from './tasks/taskConfig';
 import { RoarAppkit } from '@bdelab/roar-firekit';
 import { setTaskStore } from './taskStore';
+import { taskStore } from './taskStore';
 
 export let mediaAssets: MediaAssetsType;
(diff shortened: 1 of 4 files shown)


asengupta3:
@zwatson2001 this makes sense. I like this change. I am wondering what is the difference between `taskFinished.ts` and `finishExperiment.ts`. Should we try to put them under the same function?

asengupta3 (approved):
LGTM!

zwatson2001:
@asengupta3 great, thanks for reviewing! One of these is a jsPsych trial at the end of the task timeline if the user makes it all the way through, and the other is just content displayed on the screen if they fail/time out. I think I tried to just have one function but it was difficult because these are two different events (making it to the end of the timeline vs calling `jsPsych.endExperiment()`. I can look into it again at some point though

PR #296, opened 2025-04-04
Allow block-level randomization for theory of mind

zwatson2001:


asengupta3:
@zwatson2001 how are we going to support the scenario where some items are not randomized (like instructions) or other trials vs some which are?

task-launcher/src/tasks/shared/helpers/prepareChoices.ts
```diff
@@ -2,19 +2,52 @@ import _shuffle from 'lodash/shuffle';
...
+      choices = _shuffle(choices); 
+      break; 
+    case 'no': 
+      break; 
+    case 'at_block_level':
+      const previousChoices = taskStore().previousChoices; 
+
+      // use previous choices if they match current choices, otherwise set new choice order
+      if (
+        previousChoices != null &&
```
asengupta3, line 35: This should be `!==`.

asengupta3 (approved):
One quick comment. The rest looks good!

asengupta3:
Also please fix the conflict before merging

PR #299, opened 2025-04-11
save item IDs in memory and HAF and remove younger kid practice from SDS corpus

zwatson2001:


task-launcher/src/tasks/same-different-selection/trials/heavyInstructions.ts
kachergis, line 1: looks good, except I'm uneasy having empty itemIDs (even for instruction trials)...maybe just `instructions`?

kachergis (changes requested):
The MG and H&F itemIDs look good! Requesting a minor change to not have empty SDS itemIDs.

PR #300, opened 2025-04-16
Create cypress test for memory game

zwatson2001:
This PR adds a `memoryGameAnswer` property to the `window` object so that it can be accessed in the Cypress code.

### DIFF
--- a/task-launcher/cypress/e2e/memory.cy.js
+++ b/task-launcher/cypress/e2e/memory.cy.js
@@ -37,10 +37,10 @@ describe('test memory game', () => {
      
       if (blocks.length > 0) {
         // wait for window to contain sequence information
-        cy.window().should('have.property', 'memoryGameAnswer');
+        cy.window().its('cypressData').should('have.property', 'correctAnswer');
 
         cy.window().then((window) => {
-          const sequence = window.memoryGameAnswer;
+          const sequence = window.cypressData.correctAnswer;
           sequence.forEach((number) => {
             blocks[number].click();
           }); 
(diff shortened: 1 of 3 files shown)


task-launcher/types/index.d.ts
```diff
@@ -96,5 +96,6 @@ declare global {
   interface Window {
     Cypress: any; // FIXME: Add explict type
     initJsPsych: JsPsych;
+    memoryGameAnswer: Array<number>;
```
asengupta3, line 99: I think putting the correct answer in windows object is the right approach. Since this does not use `afcStimulus` so what you have is correct but I think we should structure this like:
```
interface Window {
    Cypress: any; // FIXME: Add explict type
    initJsPsych: JsPsych;
    cypressData: {
        correctAnswer: string | Array<number> | number;
    };
}
```
This way we can use the same variable for all the tasks? What do you think?

task-launcher/cypress/e2e/memory.cy.js
```diff
@@ -0,0 +1,74 @@
...
+    // end recursion if the task has reached the end screen
+    cy.get('p,h1').then((p) => {
+      if (p[0].textContent.includes("Thank you!")) {
+        return;
+      } else {
+        memoryLoop(); 
+      }
+    });
+  }
+
```
asengupta3, line 74: NIT: New line

task-launcher/types/index.d.ts
zwatson2001: That sounds good! Probably better than having a separate variable for each task

PR #306, opened 2025-04-17
Mental rotation downward extension

zwatson2001:
This PR adds some features to mental rotation to make it better for younger kids, including button animations and periodic encouragement trials.


CCuskley:
Do we have a testing link to send this for QA (should also probably be reviewed by psychometrics team  @mcfrank?)

PR #309, opened 2025-04-26
Fix slider instruction image and SDS OK button being cut off

zwatson2001:
This PR modifies fixes an instruction image that was not maintaining its aspect ratio.

Before:
before

After:
after


asengupta3 (approved):
LGTM!

PR #318, opened 2025-05-13
feat(308): Adding posthog/sentry logger

asengupta3:
This adds a posthog/ sentry logger to core task in a backward compatible manner. Currently it logs only data collection but can be fine tuned in the future.


zwatson2001 (approved):
Looks good to me!

PR #320, opened 2025-05-16
Math downward extension

zwatson2001:
This PR adds a few new math trial types for younger kids and runs only block 3 of the corpus (which will contain the downward extension version of math) if `heavyInstructions` is turned on.

### DIFF
--- a/task-launcher/src/tasks/math/timeline.ts
+++ b/task-launcher/src/tasks/math/timeline.ts
@@ -61,13 +61,12 @@ export default function buildMathTimeline(config: Record<string, any>, mediaAsse
 
   const { runCat, heavyInstructions } = taskStore();
 
-  if (heavyInstructions) {
-    corpus = corpus.filter(trial => {
-      return trial.block_index === "0";
-    })
+  // block 4 is only for younger kids
+  corpus = corpus.filter(trial => {
+    return heavyInstructions ? trial.block_index === "3" : trial.block_index !== "3";
+  });
 
-    taskStore('totalTrials', corpus.length);
-  }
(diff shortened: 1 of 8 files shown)


task-launcher/src/tasks/math/timeline.ts
```diff
@@ -55,11 +55,18 @@ export default function buildMathTimeline(config: Record<string, any>, mediaAsse
...
-  const corpus: StimulusType[] = taskStore().corpora.stimulus;
+  let corpus: StimulusType[] = taskStore().corpora.stimulus;
   const translations: Record<string, string> = taskStore().translations;
   const validationErrorMap: Record<string, string> = {};
 
-  const { runCat } = taskStore();
+  const { runCat, heavyInstructions } = taskStore();
+
+  // block 4 is only for younger kids
+  corpus = corpus.filter(trial => {
```
asengupta3, line 65: I am confused about the logic here. Because, with what is currently implemented it will show only block 4 when `heavyInstructions` is `true` and it will show everything apart from block 4 when `heavyInstructions` is `false`. Is that what we want?

asengupta3:
It mostly looks good to me. It will be helpful if some screenshots are posted. That helps looking at the UI while reviewing.

task-launcher/src/tasks/math/timeline.ts
zwatson2001: For right now the downward extension version of math doesn't really overlap with the existing task, so this design is a way of not having two separate corpora. I think we may want to change this in the future once we design the final version of the task, but for now I've just been differentiating between the two item sets with block number

PR #321, opened 2025-05-21
Update styling to fix instructions UI issue

zwatson2001:


asengupta3 (approved):
Can you paste some screenshots for this one as well? Ideally, one with normal screen resolution (1200x700) and another which is like (1000x433)?

But changes look good to me. Pre-approving.

zwatson2001:
Here some screenshots:
narrow_screen
normal_screen

PR #329, opened 2025-06-17
Prevent timeout interrupting the stimulus

zwatson2001:
This PR adds a check for the amount of time remaining during the fixation/setup trials and ends the task if there is not enough time to show the next stimulus. This is meant to prevent the abrupt switch to the end-of-task screen while the audio file is still playing for a trial.

### DIFF
--- a/task-launcher/src/tasks/shared/helpers/appTimer.ts
+++ b/task-launcher/src/tasks/shared/helpers/appTimer.ts
@@ -8,6 +8,8 @@ import { PageStateHandler} from './PageStateHandler';
 
 // define timerId here so that it can be cleared if task ends early
 let timerId: any;
+// buffer in milliseconds after presentation of stimulus to allow some time to answer
+const RESPONSE_BUFFER = 2000;  
 
 export const startAppTimer = (maxTimeInMinutes: number, finishExperiment: () => void) => {
   // Minimum time is 1 minute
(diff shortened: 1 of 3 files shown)


task-launcher/src/tasks/shared/helpers/appTimer.ts
```diff
@@ -1,15 +1,34 @@
...
     taskStore('maxTimeReached', true);
     finishExperiment();
     clearTimeout(timerId);
   }, maxTimeInMilliseconds);
 };
+
+// function for ending the task if the next trial
+export async function checkEndTaskEarly(timeRemaining: number, stimAudio: string) {
+    const pageStateHandler = new PageStateHandler(stimAudio, false); 
+    const minTrialDuration = await pageStateHandler.getStimulusDurationMs() + 2000;
```
asengupta3, line 28: Why the `+2000`?

zwatson2001: It's an arbitrary amount of time, but I was thinking that it should leave enough time for the participant to answer after hearing the stimulus (since cutting off exactly as the audio is ending would be kind've harsh)

asengupta3: Can we add like a variable name like `RESPONSE_BUFFER` and add that? This will make the code more readable and elicit explicitly what we are trying to do?

zwatson2001: Sure, that sounds good!

PR #335, opened 2025-07-24
Styling updates

zwatson2001:
This PR adjusts some styling to try to address the issues that Colombia is having. I resized the mental rotation buttons based on an `adjusted-viewport` variable, and removed an unnecessary top margin from the instructions. We could try resizing everything based on `adjusted-viewport`, but I went with the minimal change that might help for now.

### DIFF
--- a/task-launcher/src/styles/layout/_containers.scss
+++ b/task-launcher/src/styles/layout/_containers.scss
@@ -153,7 +153,6 @@
   width: 100%;
   max-width: 900px;
   justify-content: center;
-  margin-top: calc($replay_btn_size + $replay_btn_pos);
   margin-bottom: 16px;
   align-items: center;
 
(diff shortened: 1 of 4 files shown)


task-launcher/src/styles/layout/_containers.scss
```diff
@@ -153,7 +153,6 @@
   width: 100%;
   max-width: 900px;
   justify-content: center;
-  margin-top: calc($replay_btn_size + $replay_btn_pos);
```
asengupta3, line 156: This might lead of replay button overlap for shorter devices. Can you check if that is happening?

zwatson2001: There was some overlap on very small screens (which the original margin in place wasn't fully preventing) but I reduced the size of the replay button on a small screen to mitigate this

PR #343, opened 2025-08-15
Update gcp bucket structure

zwatson2001:
This PR allows the tasks to read assets from gcp reorganized into a different structure - rather than having a separate bucket for each task, all assets will be inside the `levante-assets-dev` bucket with [this structure](https://langcog.slack.com/archives/C05UY4Y2Q15/p1753907743735049).


digital-pro:
Aritra – I’ve just been trying to fix the issues in the PR that zach created that prevented it from being merged. So probably good to provide that feedback to him when he’s back. In the meantime I’ll pull the stress test.

digital-pro:
PS To be more clear, the problem is that at some point the cypress tests broke. So when Zach tried to deploy the new bucket structure, they failed. So fixing the cypress tests seemed necessary to deploy the new bucket structure correctly. He was headed out for a week, so I said I’d work on it. Does that makes sense?

To your point I guess we could have make a PR that fixed the bucket PR tests?

digital-pro:
I pulled out the stress test, but I left in the haf change as it is required for tests to pass.

However, there are other failing tests that I'm working on now. I'll put the fixes for those in a separate PR.

Input taken for future reference! Now if you can just get Phil not to dump 200 file PRs at 5pm on Fridays:)

PR #352, opened 2025-09-04
Revert chunk changes

asengupta3:


zwatson2001 (approved):
This all looks good to me - thanks! I don't think the tasks will work until I merge a PR that batches the asset preloading, because the changes reverted by this PR were a fix for an issue I was having where the browser was being overloaded by preloading too many assets in the beginning.

asengupta3:
Still merging this PR since it unblocks running SDS and math locally. But I agree @zwatson2001 PR would take care with the optimization

PR #364, opened 2025-09-16
Integrate downex into math cat

### LINKED ISSUE #360: Combined downward extension + older kid math CAT
The SPARK lab would like to start piloting this on 9/22 - it should start with a downward extension block that will be skipped if the user is older than 5, and then proceed to the rest of the task. A few decision points/questions (maybe for @mcfrank?):

1. How long should the first block be? This was decided based on simulation results for the other blocks, which use either 1/2 or 1/3 of the available trials. Maybe we could start with 1/2?
2. What should we use for starting IRT parameters?
3. The current math CAT design lumps together trials of a different types in a block - I think this works fine for most trials, but some of the counting trials require the assessor to answer for the child, which might be tricky if they are not together in their own block with specific instructions - could we maybe remove these?

zwatson2001:
This PR adds a block to the beginning of the math CAT with the downward extension items. The downward extension items have `block_index=3` in the corpus, so these need to be added to the beginning after the `prepareMultiBlockCat()` function splits the corpus into blocks. Closes #360

### DIFF
--- a/task-launcher/src/tasks/math/timeline.ts
+++ b/task-launcher/src/tasks/math/timeline.ts
@@ -246,7 +246,7 @@ export default function buildMathTimeline(config: Record<string, any>, mediaAsse
 
         switch (i) {
           case 0:
-            allowedIDs = ['math-instructions1-heavy'];
+            allowedIDs = ['math-instructions1-heavy', 'math-intro1-heavy'];
             break;
           case 1:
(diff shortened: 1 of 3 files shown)


asengupta3 (approved):
Changes look good. But we should look into the failing tests at some point.

CCuskley:
I don't see a testing link for this - are we okay to merge this onto Dev so that we can send it to QA?

PR #376, opened 2025-09-24
Fix cypress tests

zwatson2001:
This PR modifies the hearts and flowers cypress test so that it passes if the task is working, and also reverts some changes in the test and in H&F itself that were causing it to fail. @asengupta3 I think your recent audio changes resolved the issue with the other cypress tests.

### DIFF
--- a/task-launcher/cypress/e2e/hearts_and_flowers.cy.js
+++ b/task-launcher/cypress/e2e/hearts_and_flowers.cy.js
@@ -9,7 +9,6 @@ let mixed_test = false;
 
 describe('test hearts and flowers', () => {
   it('visits hearts and flowers and plays game', () => {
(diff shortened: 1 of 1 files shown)


task-launcher/cypress/e2e/hearts_and_flowers.cy.js
```diff
@@ -1,73 +1,39 @@
...
-};
+let heart_phase = false;
+let flower_phase = false;
+let mixed_practice = false;
+let final_instructions = false; // instructions before final mixed test phase
+let mixed_test = false;
 
 describe('test hearts and flowers', () => {
   it('visits hearts and flowers and plays game', () => {
+    cy.intercept({ resourceType: /xhr|fetch/ }, { log: false });
```
asengupta3, line 12: What does this do? Silently fails network requests?

PR #381, opened 2025-09-29
Fix math ending early

### LINKED ISSUE #380: [Bug] Math showing as complete after launch even when participant viewed no trials
When testing whether tasks launched from the dashboard (but not trying to complete them), I noticed most of them don't mark as complete if you launch it, just click a few instructions and then close it or go back - this is what we want from all the tasks I think. 

However, the math task is an exception: it shows as complete even when you just click the first couple instructions and shut it. The relevant variant in this case is the Math CAT for es-CO, but it may be a general math issue?

https://github.com/user-attachments/assets/fac74062-89be-4182-98ed-d1060af566f9

zwatson2001:
This PR fixes a bug where completion was being recorded for the math CAT early, so a user could exit out after a few trials and it would show as completed. The wrong trial counter variable was being updated in the math timeline. Closes #380

### DIFF
--- a/task-launcher/src/tasks/math/timeline.ts
+++ b/task-launcher/src/tasks/math/timeline.ts
@@ -326,8 +326,8 @@ export default function buildMathTimeline(config: Record<string, any>, mediaAsse
         fullCorpus.start.forEach((trial) => timeline.push(stimulusBlock(trial)));
       }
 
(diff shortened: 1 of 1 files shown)


asengupta3 (approved):
LGTM

PR #384, opened 2025-09-30
Fix assets-per-task.json always being read from dev

zwatson2001:
This PR fixes a bug where the assets-per-task.json file was always being read from the dev bucket, even when running on prod.


asengupta3 (approved):
LGTM!

Issue #394, opened 2025-10-30
[Task] Force instruction listening in ToM


CCuskley:
Currently, it's possible to click past the scenarious in the ToM task without listening to the full audio (see below). We should change this so that the 'OK' button below the story is inactive until the audio has finished playing, which is the current behaviour of the test trials.

https://github.com/user-attachments/assets/31f59162-cd9c-40a3-803d-b3c60e242853

PR #395, opened 2025-10-30
Disable OK button during ToM instructions

zwatson2001:
This PR disables the OK button in theory of mind while the instruction audio is playing, which should prevent spamming through the instructions. #394


asengupta3 (approved):
LGTM!

Issue #302, opened 2025-04-17
Matrix reasoning downward extension


zwatson2001:

mcfrank:
@sachinallums @SerLee will be working on making some new items for this!

mcfrank:
yes, let's wait on a concrete spec! it's part of the downward extension milestone so we should chain priority from that.

PR #397, opened 2025-11-04
Matrix reasoning downward extension

zwatson2001:
This PR adds the downward extension version of matrix reasoning. Some of the main changes:

- The audioFile field of the corpus (and StimulusType object) can now be a list of multiple audioFiles, allowing them to be timed correctly with animations that happen during the prompt. If there are multiple audio files, the text prompt is created by concatenating each translation
- The code now supports a `downex` flag in the corpus, which is used to separate the downward extension items from the older kid items
- Matrix reasoning has some new animations and styling for downward extension items
- Linked to #302


asengupta3:
@zwatson2001 Can you check the failing cypress test?

zwatson2001:
> @zwatson2001 Can you check the failing cypress test?

It looks like it was just the survey task that failed, which happened because of a missing text asset (not related to this PR). I just added the asset to dev so I think it should be fine

Issue #408, opened 2025-12-11
[Feature] Add stopping rule to matrix reasoning/pattern matching


CCuskley:
We want to end the task for users that have accuracy over the last 10 trials of <40%, i.e., they need to have gotten at least 4 of the last ten trials correct to continue. This should only kick in after 10 trails, so e.g.,

- On completion of the 10th trial, check performance over the last ten trails, if it's <4/10, end the task
- Perform the same check on every subsequent trial over window of previous 10 trials, if <4/10, end the task

PR #409, opened 2026-01-05
Stop mental rotation and matrix cat if 4 of last 10 trials are incorrect

zwatson2001:
Linked to #408

### DIFF
--- a/task-launcher/src/tasks/adult-reasoning/timeline.ts
+++ b/task-launcher/src/tasks/adult-reasoning/timeline.ts
@@ -46,7 +46,7 @@ export default function buildAdultReasoningTimeline(config: Record<string, any>,
     promptAboveButtons: true,
     task: config.task,
     layoutConfigMap,
-    catStopping: false,
+    terminateCat: false,
   };
 
   const stimulusBlock = {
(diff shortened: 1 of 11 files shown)


task-launcher/src/tasks/shared/helpers/checkCatStopping.ts
```diff
@@ -0,0 +1,12 @@
+import { jsPsych } from "../../taskSetup";
+import { finishExperiment } from "../trials/finishExperiment";
+
+// ends the task if 4 of the last 10 trials have been incorrect
+export function checkCatStopping() {
+    const data = jsPsych.data.get().filter({assessment_stage: 'test_response'}).last(10);
```
asengupta3: Do you think we need to filter anything else other than `test_response`? I haven't looked at the corpuses in a while and am wondering if there are others we introduced that don't count in checking for correctness.

task-launcher/src/tasks/shared/helpers/checkCatStopping.ts
```diff
@@ -0,0 +1,12 @@
+import { jsPsych } from "../../taskSetup";
+import { finishExperiment } from "../trials/finishExperiment";
+
+// ends the task if 4 of the last 10 trials have been incorrect
+export function checkCatStopping() {
```
asengupta3: rename to `shouldTeminateCat`?

PR #410, opened 2026-01-09
Downex updates

zwatson2001:
This PR addresses the feedback we received on the new downward extension tasks (which is written out [here](https://www.notion.so/Notes-from-Kali-re-Downex-2c7244e26d9b8078b132e3b97d63dac9)). The changes it introduces are: 

- Preventing audio overlap between trials in matrix reasoning with animation/audio sequences
- Fixing a bug that prevented audio from playing on iOS without user interaction (the audio context is now unlocked when the full screen button is pressed at the beginning of the task)
- Modified the animations of the downex memory game trials so the blocks light up for longer after being clicked
- Set `border-radius: 0` on images in matrix reasoning so that their edges are not cut off

### DIFF
--- a/task-launcher/src/tasks/memory-game/trials/stimulus.ts
+++ b/task-launcher/src/tasks/memory-game/trials/stimulus.ts
@@ -28,6 +28,9 @@ let sequenceLength = 2;
 let generatedSequence: number[] | null;
 let selectedCoordinates: [number, number][] = [];
 let numCorrect = 0;
+const HIGHLIGHT_COLOR = '#8CAEDF';
+const INCORRECT_COLOR = '#f00';
+
 
 // store the finish trial function here (we have to override it later to prevent auto-finish on display trials)
 const originalFinishTrial = jsPsych.finishTrial;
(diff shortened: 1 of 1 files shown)


task-launcher/src/tasks/shared/helpers/audioHandler.ts
```diff
@@ -64,4 +64,20 @@ export class PageAudioHandler {
       return;
     }
   }
+
+  // required on iOS to prevent autoplay blocking
+  static unlockAudioContext() {
+    const ctx = jsPsych.pluginAPI.audioContext();
+
+    // safari requires resuming audio context on user interaction, then it can be used freely later
+    if (ctx.state === 'suspended') {
```
asengupta3: Wow! This is great find! Great job @zwatson2001

asengupta3 (approved):
These are really good changes. I have not tested them but the code looks good!

task-launcher/src/tasks/shared/helpers/audioHandler.ts
zwatson2001: Thanks! I'm hoping this will fix those intermittent audio problems we were seeing on Safari a while ago

PR #411, opened 2026-01-22
add demo mode (no Firestore)

zwatson2001:
This PR adds a "demo mode" to core tasks, which turns of all interactions with Firestore. This is necessary because if the tasks are run by a user that isn't logged in (like on the QA link or locally) some of the writes to the Firestore database are failing due to insufficient permissions. Demo mode is enabled by default when running locally (added to `npm run dev` and `npm run dev:db`), and would also be enabled when deploying to the QA link.


asengupta3 (commented):
So we just don't write trials in demo mode? Do we want to at least log it to console, maybe with a filter?

asengupta3 (approved):
LGTM

PR #433, opened 2026-04-07
Sds updates

### LINKED ISSUE #386: Something's the Same Spark Updates
Something's the same task needs to be modified for use with younger children.

- Age dependent starting logic? (Planning doc does not say this, but should check)
- Changes to framing and terminology which will include use of the concept of "match" between two items
- Potential addition of criterion trials which show 4 cards with two pairs to evaluate understanding of "match"
    - To check: For all users or only younger children?

This will require:
- Changes to the item bank
- Potentially new images/illustration?
- Changes to translations in Spanish and German at minimum, plus any other active languages.

The current state of the spec for this is in [Notion](https://www.notion.so/Downward-extension-planning-9-25-272244e26d9b802488e9c2eedcd9bc6e?source=copy_link) - this issue should remain in the backlog until there is a full spec, and may require the use of sub-issues.

zwatson2001:
This PR makes a few small updates to SDS, adding an instruction screen before the first 5-card matching trial and preventing the user from continuing beyond a practice trial until they get it right.

### DIFF
--- a/task-launcher/src/tasks/same-different-selection/catTimeline.ts
+++ b/task-launcher/src/tasks/same-different-selection/catTimeline.ts
@@ -100,7 +100,7 @@ export default function buildSameDifferentTimelineCat(config: Record<string, any
         timeline.push(buttonNoise);
       }
 
-      if (i < trialNum - 1) {
+      if (i < trialNum - 1 && !trialType.includes('something-same')) {
         timeline.push({ ...fixationOnly, stimulus: '' });
       }
     }
--- a/task-launcher/src/tasks/same-different-selection/catTimeline.ts
+++ b/task-launcher/src/tasks/same-different-selection/catTimeline.ts
@@ -99,10 +99,6 @@ export default function buildSameDifferentTimelineCat(config: Record<string, any
         timeline.push(afcMatch());
         timeline.push(buttonNoise);
       }
-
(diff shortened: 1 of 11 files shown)


task-launcher/src/tasks/same-different-selection/catTimeline.ts
```diff
@@ -125,7 +121,23 @@ export default function buildSameDifferentTimelineCat(config: Record<string, any
   const timeline = [preloadTrials, initialTimeline];
 
   // all instructions + practice trials
-  const instructionPractice: StimulusType[] = heavy ? preparedCorpus.ipHeavy : preparedCorpus.ipLight;
+  let instructionPractice: StimulusType[] = preparedCorpus.ipLight;
+
+  let fiveBlockIntroTrial: StimulusType;
+  let fiveBlockIntro: any; 
+  if (taskStore().taskVersion === 2) {
```
asengupta3, line 128: Is there a place where `taskVersion` is set to 2. Is there a way we manage it automatically?

zwatson2001: It's a url/variant param, so it's either set in the variant doc or passed in through the url if it's being run with the demo link. The plan is to set it to 2 for only the language variants that have translations for the new strings so that we can deploy these changes one language at a time.

asengupta3 (approved):
LGTM

PR #442, opened 2026-04-17
Hearts and flowers redesign

### LINKED ISSUE #412: Adaptive hearts and flowers timing
Right now H&F is untimed. We want to move it to fixed timing at 2s. 

But 2s is too short for the oldest kids and too long for the youngest. One possible solution is to use some combination of age-dependent timing and block-adaptive timing. 

Here is a very simple idea: 

Everyone gets 2s hearts block, 2s flowers block, 2s mixed block (but we shorten the mixed block). Then we give another block based on jsCat ability estimates from the rest of the task, chosen from these options (from easiest to hardest):
- 3s flowers block 
- 2s flowers block
- 3s mixed block
- 2s mixed block
- 1.5s mixed block

Pro of this option is that we already have the parameters for everything except the new blocks and this requires no age-dependent timing. We just need to associate basic CAT parameters to the items and run jsCat, then add another block. We also can fit models to the new blocks because we are guaranteed to have overlap with the blocks we already have. 

Con is that maybe we want to make the task slower for all blocks for the youngest kids. That would mean that we need to figure out how to give slower hearts and flowers tasks. The trouble is that we would then have no item overlap and no ability to fit models to the downex and regular data together. 

Thoughts @mikabr @CCuskley @zwatson2001 ?

### LINKED ISSUE #439: Hearts and flowers overhaul
Hearts and flowers overhaul based on slack doc

zwatson2001:
Implements changes to hearts and flowers described here: https://stanford.enterprise.slack.com/docs/T052J7XMP/F0AMLQBRM8V


asengupta3 (approved):
LGTM! Did not test but the code looks good. We probably will need to manage our eventListeners a little better at some point. We are adding a bunch of them but never removing them. This can lead to degraded performance over time.

PR #445, opened 2026-04-21
Add styling to fix button formatting issues on older ipads

zwatson2001:
This PR introduces a small change that fixes the text centering issue happening on some older ipads. I verified that the formatting is correct on one of the ipads with this change.


asengupta3 (approved):
LGTM!

PR #447, opened 2026-04-27
Theory of mind random story selection

zwatson2001:
This PR updates theory of mind according to this spec: https://stanford.enterprise.slack.com/docs/T052J7XMP/F0AV91LMV7W. The task should now show three stories. Each story is randomly selected from a pool of a certain kind of item. This is done during the task at the end of each story rather than in advance so that this can be easily changed to a CAT in the future.

Because the stories are made up of individual items that have to stay in order, this PR takes the first item in each story and puts those into `taskStore().corpora.stimulus`. The rest are put into `taskStore().sequentialTrials`. In between stories, a random item with the right `storyGroup` (defined in corpus) is selected from `taskStore().corpora.stimulus`. During stories, the next items from that story are selected from `taskStore().sequentialTrials`.


asengupta3 (approved):
LGTM! Looks like one of the tests is failing consistently. Can you fix that before you merge?

PR #448, opened 2026-05-01
Bugfix: blockNumber === 0

u2234-io:
- Change `task-launcher/src/tasks/shared/helpers/getStimulus.ts` L17 to properly evaluate to truthy when `blockNumber === 0`
  - NB: `!= null` is preferred here in order to evaluate both `undefined` and `null` to falsy
- Change L20 and L74 for consistency (they weren't bugged, but avoids future readers wondering why those conditionals are written differently than L17)


CCuskley (approved):
LGTM

PR #450, opened 2026-05-05
Switch to 3-parameter CAT

zwatson2001:
This PR incorporates the discrimination value of each item into the theta estimation function now that it is available in the corpus.


CCuskley:
The code for this looks okay, but I'm concerned about the failing tests for es-AR- the de-DE matrix reasoning failure is a known issue with the first asset loading timing out - but other tests?

PR #452, opened 2026-05-06
Theory of mind fake CAT updates

zwatson2001:
Updates the default number of stories in the task to 4 and adds a fallback for any gender/number specific audio assets ending in -fem or -plural.


zwatson2001:
These changes are no longer needed, so I'm closing the PR

PR #453, opened 2026-05-11
Prevent failure if lng='en-US' and add timing fallback to memory game

zwatson2001:
Adds more robustness to language checks, preventing failure when the language is set to 'en-US' rather than the legacy 'en'.


u2234-io:
`taskStore().language.includes('en')` returns false positive on codes like `"ven"` (Venetian) and `"ben"` (Bengali). We don't currently support these languages, but we might as well make these checks use a more robust helper while we're touching them. It should go somewhere in `task-launcher/src/utils`.

```ts
// Minimal one-liner
const isEnglish = (lang: string) => lang === 'en' || lang.startsWith('en-');

// Or use the Intl API for robust BCP 47 parsing (supported by all browsers since 2019/20)
const isEnglish = (lang: string) => {
  try {
    return new Intl.Locale(lang).language === 'en';
  } catch {
    return false;
  }
};
```

zwatson2001:
> `taskStore().language.includes('en')` returns false positive on codes like `"ven"` (Venetian) and `"ben"` (Bengali). We don't currently support these languages, but we might as well make these checks use a more robust helper while we're touching them. It should go somewhere in `task-launcher/src/utils`.
> 
> ```ts
> // Minimal one-liner
> const isEnglish = (lang: string) => lang === 'en' || lang.startsWith('en-');
> 
> // Or use the Intl API for robust BCP 47 parsing (supported by all browsers since 2019/20)
> const isEnglish = (lang: string) => {
>   try {
>     return new Intl.Locale(lang).language === 'en';
>   } catch {
>     return false;
>   }
> };
> ```

Good catch, thanks! I just made this change.

u2234-io (approved):
LGTM. Thank you!

PR #454, opened 2026-05-12
Add experimenter buttons

zwatson2001:
This PR adds two buttons to the tasks designed to be used by the experimenter supervising the child: exit and pause. It also adds a fullscreen button that appears if the task is ever not in fullscreen. 

https://github.com/user-attachments/assets/d8f4e0ed-57ab-4cc9-af0c-2f1985a88f05


asengupta3 (approved):
LGTM! I did not test this code locally but the code looks good. We should probably design our architecture in a way where these changes do not need to touch this many files. Like maybe have a base component and then the each task inherit from that initial page setup. We can discuss this when we meet at Stanford @zwatson2001

task-launcher/src/tasks/matrix-reasoning/trials/downexStimulus.ts
```diff
@@ -16,7 +18,7 @@ import { isTouchScreen, jsPsych } from '../../taskSetup';
 const replayButtonHtmlId = 'replay-btn-revisited';
 let practiceResponses = [];
 let startTime: number;
-let audioEnabled = false; // disable audio if the trial has changed since the loop started - prevent overlapping audio
+let cycleId = 0; // disable audio if the trial has changed since the loop started - prevent overlapping audio
```
asengupta3, line 21: What does the `cycleId` do?

zwatson2001:
> LGTM! I did not test this code locally but the code looks good. We should probably design our architecture in a way where these changes do not need to touch this many files. Like maybe have a base component and then the each task inherit from that initial page setup. We can discuss this when we meet at Stanford @zwatson2001

Thanks! Yeah that sounds good

task-launcher/src/tasks/matrix-reasoning/trials/downexStimulus.ts
zwatson2001: It's there to keep track of the how many times the animation sequence was triggered for the downex matrix reasoning trials (which is needed because it can now be triggered from multiple sources, like unpausing the task or using the replay button). Without it the audio and animations can overlap if the user pauses/unpauses multiple times

PR #457, opened 2026-05-18
Increase time limit to prevent asset-heavy tasks from timing out

zwatson2001:


asengupta3 (approved):
LGTM! Do you know why it is taking so much time to load?

zwatson2001:
> LGTM! Do you know why it is taking so much time to load?

Thanks for reviewing! I think it's just because this task has more images than any other task. The reason it was happening for DE-de and not es-CO is just because the German test ran first, and the images are already cached when the Spanish one runs.

PR #460, opened 2026-05-22
Hearts and flowers updates

zwatson2001:
This PR makes the updates to hearts and flowers request in this [document](https://docs.google.com/document/d/1eTbX2tgxnAOJIwZ0qZGLZCHPhhok6MNlcdmQdyvD10c/edit?usp=sharing). The demo that lets the child practice using the buttons is now broken up into multiple trials, and the "press the spacebar to continue" text now has accompanying audio for desktop users. For touch screen users, the toast warning against using a mouse is disabled and the ok button to advance to the next trial is not enabled until the instructions are finished playng.

### DIFF
--- a/task-launcher/src/tasks/hearts-and-flowers/trials/instructions.js
+++ b/task-launcher/src/tasks/hearts-and-flowers/trials/instructions.js
@@ -14,6 +14,7 @@ import { disableOkButton } from '../../shared/helpers/disableOkButton';
 import { enableOkButton } from '../../shared/helpers/enableButtons';
 
 let continueTrialConfig;
+let cleanupInstructionInputListeners = [];
 
 function isHfV2() {
   return taskStore().version === 2;
@@ -77,7 +78,6 @@ function buildInstructionTrial(mascotImage, getPromptKey, showResponseButton = f
   }
 
   const replayButtonHtmlId = 'replay-btn-revisited';
-  let cleanupInstructionInputListeners = null;
 
   const trial = {
     type: jsPsychHtmlMultiResponse,
@@ -184,12 +184,16 @@ function buildInstructionTrial(mascotImage, getPromptKey, showResponseButton = f
                   maxRepetitions: 2
                 },
                 onEnded: () => {
-                  window.addEventListener('keydown', (event) => {
+                  const onSpacebarPress = (event) => {
                     if (event.key === " ") {
                       jsPsych.finishTrial();
                     }
-                  }, 
-                  {once: true});
+                  };
+      
+                  window.addEventListener('keydown', onSpacebarPress);
+                  cleanupInstructionInputListeners.push(() => {
+                    window.removeEventListener('keydown', onSpacebarPress);
+                  });
                 }
               }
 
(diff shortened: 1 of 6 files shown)


task-launcher/src/tasks/hearts-and-flowers/trials/instructions.js
```diff
@@ -129,10 +136,10 @@ function buildInstructionTrial(mascotImage, getPromptKey, showResponseButtons =
         buttonContainer.classList.add('linear-4');
         buttonContainer.innerHTML = `
           <div class='response-container--small'>
-            <button class='secondary--green'></button>
+            <button class='secondary--green' ${buttonSide === "right" ? 'style=visibility: hidden disabled' : ''}></button>
```
u2234-io: `visibility: hidden` needs to be wrapped with double quotes: `<button class='secondary--green' ${buttonSide === "right" ? 'style="visibility: hidden" disabled' : ''}></button>`

Same thing on line 142 just below.

task-launcher/src/tasks/hearts-and-flowers/trials/instructions.js
```diff
@@ -220,7 +227,8 @@ function buildInstructionTrial(mascotImage, getPromptKey, showResponseButtons =
         },
       };
 
-      PageAudioHandler.playAudio(mediaAssets.audio[getPromptKey()] || mediaAssets.audio.inputAudioCue, audioConfig);
+      const promptAudioKey = showResponseButton ? getPromptKey(true) : getPromptKey(false);
+      PageAudioHandler.playAudio(mediaAssets.audio[promptAudioKey] || mediaAssets.audio.inputAudioCue, audioConfig);
 
       const pageStateHandler = new PageStateHandler(getPromptKey());
```
u2234-io: I think this may need to be `const pageStateHandler = new PageStateHandler(promptAudioKey);` so that it always replays the same audio it did on line 231 when `showResponseButton` is true.

`getPromptKey()` is equivalent to `getPromptKey(false)` so it would work as expected for that case, but not for the `getPromptKey(true)` case.

Probably worth testing before & after to confirm behavior tho. Maybe I'm misunderstanding.

u2234-io:
We should verify the following translation/audio keys exist:
- `heartsAndFlowersInstructButtonTouch`
- `heartsAndFlowersInstructKeyPress`
- `heartsAndFlowersPressAnyKey`

`cypress/e2e/hearts_and_flowers.cy.js` currently only runs the default (v1) flow and clicks OK through instructions. So it won't exercise:
- Split left/right button demos
- Space-only continue after dual audio
- Touch-specific paths

Consider adding a H&F v2 Cypress run. I think that would cover the asset verification above?

task-launcher/src/tasks/hearts-and-flowers/trials/instructions.js
```diff
@@ -144,74 +151,74 @@ function buildInstructionTrial(mascotImage, getPromptKey, showResponseButtons =
...
+                  enabled: false,
+                  maxRepetitions: 2
+                },
+                onEnded: () => {
+                  window.addEventListener('keydown', (event) => {
+                    if (event.key === " ") {
+                      jsPsych.finishTrial();
+                    }
+                  }, 
+                  {once: true});
```
u2234-io: `once: true` is a bit fragile for this block. If for some reason the trial ends without this executing then the event handler lingers and could cause an exception.

I think you can use the same pattern as on line 223 below:

```
onEnded: () => {
  const onSpaceKeydown = (event) => {
    if (event.key === ' ') {
      jsPsych.finishTrial();
    }
  };
  window.addEventListener('keydown', onSpaceKeydown);
  cleanupInstructionInputListeners = () => {
    window.removeEventListener('keydown', onSpaceKeydown);
  };
},
```

task-launcher/src/tasks/hearts-and-flowers/trials/instructions.js
```diff
@@ -144,74 +151,75 @@ function buildInstructionTrial(mascotImage, getPromptKey, showResponseButtons =
...
-                { once: true },
-              );
-            });
+            const buttonPressListener = (event) => {
+              onButtonPress(displayedButton, displayedButtonIndex, event);
+            }
+
+            displayedButton.addEventListener('touchend', buttonPressListener);
+            cleanupInstructionInputListeners = () => {
+              displayedButton.removeEventListener('touchend', onButtonPress);
```
u2234-io: Should be `displayedButton.removeEventListener('touchend', buttonPressListener);`

task-launcher/src/tasks/hearts-and-flowers/trials/instructions.js
u2234-io: I still think this is worth improving.

u2234-io (approved):
LGTM

Issue #466, opened 2026-06-09
Fix linter


u2234-io:
The `lint` script in `task-launcher/package.json` is currently misconfigured. The primary issue is that it's missing the eslint CLI's `[file|dir|glob]` positional arg. Therefore it silently executes against 0 files and speciously reports 0 errors.

Adding the missing arg, i.e., `npx eslint --ext .ts,.tsx,.js,.jsx .`, it correctly returns 610 errors and 19 warnings.

Some of these are false errors stemming from the fact that eslint isn't properly configured for Typescript, but many of them are genuine.

If we stick w/ eslint, then we need to
1) install and configure `typescript-eslint`
2) fix the script, e.g., `eslint . --ext .ts,.js --max-warnings 0`
3) fix the remaining errors

---

Alternatively, I prefer `@biomejs/biome` for linting/formatting because it accomplishes the same as eslint+prettier w/ far fewer dependencies and far less configuration, and runs ~10-25x faster on benchmarks. Since we're already going to be authoring a large PR to fix the current linting errors, we may consider the switch. I wouldn't suggest it if the status quo wasn't broken.

Benchmark | Biome | ESLint+Prettier
-- | -- | --
Linting speed (large monorepo) | ~0.2-3.2 s | ~3-94 s
Formatting (1,000+ files) | ~50 ms | ~1-2 s
Cold starts (CI/CD) | Instant (3-5x improvement) | Node.js overhead
Installation size | ~10 MB binary | 100+ dependencies

digital-pro:
That makes sense to me. What’s the best way to “memorialize” these preferences? Obviously Notion docs are a thing, when we remember to read them. Any upside to stuffing them into a .rules file or similar?

From: u2234-io ***@***.***>
Sent: Tuesday, June 9, 2026 3:52 PM
To: levante-framework/core-tasks ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [levante-framework/core-tasks] Fix linter (Issue #466)

[...]

PR #470, opened 2026-06-19
Switch eslint+prettier to biome

u2234-io:
This PR replaces eslint+prettier tooling w/ biome as per discussion here: #466.

Notes:
- Biome is configured to match the previous prettier config as closely as possible to reduce commit noise.
- CI workflows only run the formatter check currently. We will need to enable the linter check as well once the errors are fixed. I left TODO comments so we don't forget.

---

After the change, here's the current linter error count:

```
Found 134 errors.
Found 228 warnings.
Found 31 infos.
```

Many of these can likely be auto-fixed, but that should be done in a separate PR.


hdlee8577 (approved):
Pulled the branch and checked locally:
- 'npm run format:check' passes
- 'npm run check' reports 134 errors (matches the PR description)
- Diff looks tooling-only — no logic changes in src/

Looks good to me, approving.

PR #471, opened 2026-06-22
fix: apply biome check:fix auto-fixes (#466)

### LINKED ISSUE #466: Fix linter

hdlee8577:
Related to #466
**Summary**
- Applies `npm run check:fix` auto-fixes across task-launcher
- Restores `@ts-ignore` placement in 3 files broken by organizeImports (build was failing with 7 TS errors)

**Lint**
- Before: 134 errors, 228 warnings
- After: 74 errors, 183 warnings

**Changes**
Mostly import organization, `let` → `const`, unused vars, and formatting. No intentional logic changes. Reviewed diff manually.

**Verification**
- [v] `npm run format:check` passes
- [v] `npm run build` passes (3 webpack warnings, same as main)


hdlee8577:
Biome auto-fix turned .map(...).flat() → .flatMap() in the matrix-reasoning downex fallback block (2 spots). Functionally the same refactor, but it was the only behavior-adjacent change in this PR, so I reverted it. Rest of the diff is formatting only.

zwatson2001 (approved):
Looks good to me, thanks! The PRs I just merged caused some conflicts, so I'm happy to resolve those if that would be helpful.

PR #477, opened 2026-06-26
Bug fixes for jun release

zwatson2001:
This PR fixes two minor bugs in mental rotation. The practice transition screen ("Good job with the practice! Let's try some more on your own.") is now only added to the timeline if there are practice trials in that block. It also fixes a bug that was causing the repeat instructions feature not to run at the right time in the CAT version and improves the responsiveness of the child survey buttons to smaller screen sizes.

### DIFF
--- a/task-launcher/src/styles/abstracts/_mixins.scss
+++ b/task-launcher/src/styles/abstracts/_mixins.scss
@@ -11,10 +11,10 @@ $breakpoints: (
   'large': (
     max-width: 1400px,
   ),
-  'very-large': (
+  'large-lower-bound': (
     min-width: 1000px,
   ),
-  'huge': (
+  'very-large-lower-bound': (
     min-width: 1200px,
   ),
 );
--- a/task-launcher/src/styles/abstracts/_mixins.scss
+++ b/task-launcher/src/styles/abstracts/_mixins.scss
@@ -9,7 +9,7 @@ $breakpoints: (
     max-width: 800px,
   ),
   'large': (
-    max-width: 1000px,
+    max-width: 1400px,
   ),
   'very-large': (
     min-width: 1000px,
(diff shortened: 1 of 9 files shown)


CCuskley:
This seems to be failing a test related to doing the child survey - @zwatson2001 is the test misconfigured or is there something wrong with the task?

zwatson2001:
> This seems to be failing a test related to doing the child survey - @zwatson2001 is the test misconfigured or is there something wrong with the task?

It was a bug in the test rather than the task - I just updated the test so hopefully it should pass now.

task-launcher/src/styles/abstracts/_mixins.scss
```diff
@@ -9,7 +9,7 @@ $breakpoints: (
     max-width: 800px,
   ),
   'large': (
-    max-width: 1000px,
+    max-width: 1400px,
   ),
   'very-large': (
     min-width: 1000px,
```
digital-pro, line 15: Probably doesn't matter, but should very-large now start at 1400?

digital-pro (approved):

task-launcher/src/styles/abstracts/_mixins.scss
zwatson2001: Oh good catch, thanks! I don't think it's used but I'll update it for consistency

task-launcher/src/tasks/mental-rotation/catTimeline.ts
```diff
@@ -193,24 +195,21 @@ export default function buildMentalRotationCatTimeline(config: Record<string, an
...
-
       // push in starting block
-      corpora.start.forEach((trial: StimulusType) => {
+      const fallBackIndex = 4;
+      corpora.start.forEach((trial: StimulusType, i: number) => {
         timeline.push({ ...fixationOnly, stimulus: '' });
         timeline.push(afcStimulusTemplate(trialConfig, trial));
         timeline.push(ifRealTrialResponse);
+
+        if (i < fallBackIndex) {
```
u2234-io, line 205: Possible that this is an intentional change, but should this be `i <= fallBackIndex`? That's what it was on the deleted L211 below.

As is, the fallback check will happen after the first 4 items, but not after the 5th. So if the user gets the first 3 items correct but the next 2 wrong they won't be shown the `fallbackInstructions` block.

u2234-io (approved):

PR #467, opened 2026-06-09
CI: npm OIDC

u2234-io:
#465

- Change `deploy-and-publish.yml` to use npm OIDC instead of long-lived token
- Other miscellaneous `deploy-and-publish.yml` hardening/cleanup:
  - Add `concurrency` block
  - Add `permissions` block
  - Set `working-directory` default
  - Use `actions/create-github-app-token@v3` over deprecated `tibdex/github-app-token@v2`
  - Bump `actions/checkout`, `actions/setup-node` to v6
  - Migrate lint/format checks from implicit `preversion`/`version`/`postversion` scripts
  - Update inputs context access for consistency
- Bump Node to 24.x (required for OIDC)
  - Pinned w/ `.nvmrc` & `engines` field in `package.json`
- Remove unnecessary `npx` calls from `package.json`
- Add missing `repository.url` to `package.json`

PR #482, opened 2026-06-30
Chore: bump to 1.3.4

u2234-io:
The previous run of `deploy-and-publish.yml` revealed an issue that I introduced in #467 - it published 1.3.4 to npm, but failed to push the bump commit/tag. This occurred because `npm version` does not create refs when run in a subdirectory.

This PR manually bumps `task-launcher/package.json`, `task-launcher/package-lock.json` to 1.3.4 and fixes the issue in `.github/workflows/deploy-and-publish.yml`.

Also adds an `Ensure on main` guard step, and reorders the release flow so npm publish happens before git commit/tag (in order that the `main`/`module` values in `package.json` no longer drift).

### DIFF
--- a/.github/workflows/deploy-and-publish.yml
+++ b/.github/workflows/deploy-and-publish.yml
@@ -52,6 +52,13 @@ jobs:
         with:
           token: ${{ steps.checkout_token.outputs.token }}
 
+      - name: Ensure on main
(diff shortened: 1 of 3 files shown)


zwatson2001 (approved):
LGTM. Thanks!

PR #483, opened 2026-07-01
Add individual replay buttons to survey responses

zwatson2001:
This PR adds replay buttons to each survey response button that can be used to replay the audio for the corresponding button. It also updates the styling to add a 'selected' state that is more consistent with existing styles. 

Before changes: 

https://github.com/user-attachments/assets/e6162937-89b4-4373-baf9-f127a9001893

After changes: 

https://github.com/user-attachments/assets/f9e62034-3ae1-4695-9760-dee02e62e771


asengupta3:
One of the languages test is failing. Can you please look into it?

asengupta3:
I wonder if the replay button should be a part of the button setup itself? Also, have you tested how does it look for smaller screens now?

zwatson2001:
> One of the languages test is failing. Can you please look into it?

There was an issue with the pt-BR math file, but it's resolved now (I think the cypress tests are passing).

asengupta3 (approved):
LGTM!

PR #488, opened 2026-07-07
Improve responsiveness of sds response buttons and add touch listeners

zwatson2001:
This PR addresses two problems: one is an issue reported by a user where the response buttons in same different selection were not responsive at all on an ipad. I wasn't able to reproduce this, but I added separate touch listeners to try to solve this problem. The other is that the buttons were not responding immediately after the trial begins - this is because the jsPsych plugin disables all response buttons as soon as the first one is clicked. Previously, these buttons were reenabled after a 500 ms delay, but this PR continuously polls for the moment when the buttons are disabled and then reenables them, allowing them to be enabled sooner.


task-launcher/src/tasks/same-different-selection/trials/afcMatch.ts
```diff
@@ -335,9 +340,24 @@ export const afcMatch = (trial?: StimulusType) => {
...
-            setTimeout(() => enableBtns(responseBtns), 500);
-          }),
-        );
+            if (firstClick) {
+              await isTaskFinished(() => card.disabled, 10).then(() => enableBtns(responseBtns));
+              firstClick = false;
+            }
+          };
+
+          card.addEventListener('touchend', () => {
```
asengupta3: I wonder if this is the root cause, then should we create a wrapper function which does the same thing when an element is either touched and/ or clicked. That way we do not have to do this every single time for every single element.

zwatson2001: That makes sense - I added a helper function that could handle this for any button. Let me know if this is what you had in mind!

asengupta3 (commented):
LGTM! Looks like there are some merge conflicts and lint errors. Once fixed, you can merge whenever.

asengupta3 (approved):
LGTM!

PR #489, opened 2026-07-08
Remove unused or hardcoded strings

zwatson2001:
This PR removes two strings and replaces them with strings that have duplicate text (the unused ones are being hidden in Crowdin):

- memoryGameForwardPrompt -> memoryGameInput
- matrixReasoningInstruct1Part2 -> matrixReasoningPrompt1Part2

It also adds general-yes and general-no, which need to be added to Crowdin, and removes some hardcoded words or fallbacks to English.


CCuskley (approved):
LGTM - though there are some minor lint errors. Try to fix those before merge

PR #490, opened 2026-07-12
Fix mislabeled 'Task Aborted' completion event in finishExperiment

digital-pro:
## Summary

`finishExperiment()` runs at the **normal end of a task** — it is the shared terminator that renders the "Task Finished!" screen and calls `jsPsych.endExperiment(...)`. Despite that, it captured a PostHog event named **`Task Aborted`**. As a result, every task whose termination path is `finishExperiment` reported a successful completion as an *abort*.

This PR renames that event to **`Task Finished`**, matching the event already emitted by the other terminal path (`taskFinished.ts`), so every task reports a single, correctly-named completion event.

The change is intentionally minimal: **the net diff is a one-line string change.** No control flow, DOM, listeners, timeline, audio, or Firestore writes are touched. Unifying the two terminal functions (`finishExperiment` vs `taskFinished`) is deliberately left for a separate follow-up.

## Background / why this matters

There are two ways a task ends:

- **`taskFinished()`** — a jsPsych **trial** appended to the end of the timeline. Fires `Task Finished`. Only reached if the timeline runs to its natural end (e.g. theory-of-mind, intro, child-survey).
- **`finishExperiment()`** — an **imperative terminator** called mid-timeline (corpus exhausted, CAT stop via `shouldTerminateCat`, app timeout via `checkEndTaskEarly`, data-quality stop via `dataQuality`). It calls `jsPsych.endExperiment(...)`, which short-circuits the remaining timeline — so the trailing `taskFinished()` trial is **never reached** for these tasks.

Because adaptive / corpus-limited tasks almost always end via `finishExperiment`, their only end-of-task event was the mislabeled `Task Aborted`. This produced badly misleading analytics.

### Evidence (PostHog, prod, last 30 days)

| task | Task Launched | Task **Aborted** | Task **Finished** | terminal path |
| --- | --- | --- | --- | --- |
| same-different-selection | 656 | **644 (98%)** | 5 | `finishExperiment` |
| memory-game | 70 | 36 | 0 | `finishExperiment` |
| mental-rotation | 72 | 39 | 6 | `finishExperiment` |
| egma-math | 69 | 38 | 20 | `finishExperiment` |
| matrix-reasoning | 46 | 36 | 0 | `finishExperiment` |
| vocab | 42 | 29 | 4 | `finishExperiment` |
| trog | 38 | 29 | 2 | `finishExperiment` |
| theory-of-mind | 623 | 9 | 588 (94%) | `taskFinished` |
| intro | 612 | 0 | 603 | `taskFinished` |

Same-Different-Selection showed a ~98% "abort" rate that is actually a ~95% **completion** rate: a Firestore check of real-site SDS runs in `hs-levante-admin-prod` found 1,124/1,227 completed (91.6%) for 2026 and 633/665 (95.2%) in the last 30 days, with only ~5.5% empty runs. The "abort" signal was an artifact of the event name, not real abandonment.

## Change

`task-launcher/src/tasks/shared/trials/finishExperiment.ts`:

```diff
-    logger.capture('Task Aborted', {
+    logger.capture('Task Finished', {
       taskName: taskStore().task,
       taskFinished: taskStore().taskComplete,
     });
```

(The branch contains two commits — remove, then re-add under the correct name — but the net effect vs `main` is the single line above.)

## Safety analysis

- **No functional/runtime risk.** Pure telemetry string; task execution, data capture, and run completion (`completed` flag, written by the dashboard) are unaffected.
- **No double-counting.** The two terminal paths are mutually exclusive per run (`endExperiment` kills the trailing `taskFinished()` trial). Verified in data: grouping PostHog events by session **and** task, only 3 of ~2,155 (session, task) pairs emit both — and those are same-task retries within one sitting.
- **No broken downstream assets.** No saved PostHog insight, dashboard, or action references `Task Aborted` or `Task Finished`.

## Expected analytics impact (not a bug)

- `Task Aborted` will drop to ~0.
- `Task Finished` volume steps **up** by roughly the former `Task Aborted` volume (~+982/30d) as the `finishExperiment`-terminal tasks now report under it. Anyone trending either event will see a discontinuity at the release point — recommend a PostHog annotation when this ships.
- Pre-existing quirk (unchanged, inherited from `taskFinished.ts`): the captured `taskFinished` property is read before the user clicks Exit, so it is ~always `false`; don't use it as a completion flag.

## Follow-up (out of scope here)

Unify `finishExperiment` and `taskFinished` behind a single shared end-screen renderer (DOM/audio/i18n/listeners + one completion event with a `reason`/`status`), keeping both control-flow entry points and the completed-vs-terminated-early distinction. This needs per-task QA (fullscreen exit, end-screen audio, no stray-keypress dismissal, data-row parity).

## Test plan

- [ ] Run a `finishExperiment`-terminal task (e.g. same-different-selection) to completion and confirm a single `Task Finished` event fires (and no `Task Aborted`).
- [ ] Run a `taskFinished`-terminal task (e.g. theory-of-mind) and confirm `Task Finished` still fires exactly once.
- [ ] Trigger an early-termination path (e.g. app timeout / data-quality stop) and confirm `Task Finished` fires and the run still records trials + `completed` in Firestore.
- [ ] Confirm the end screen, Exit button, and end-of-task audio still behave as before.

### DIFF
--- a/task-launcher/src/tasks/shared/trials/finishExperiment.ts
+++ b/task-launcher/src/tasks/shared/trials/finishExperiment.ts
@@ -2,6 +2,7 @@ import { jsPsych } from '../../taskSetup';
 import { PageAudioHandler } from '../helpers';
 import { mediaAssets } from '../../..';
 import { taskStore } from '../../../taskStore';
+import { Logger } from '../../../utils';
 
 export function finishExperiment() {
   const t = taskStore().translations;
@@ -24,6 +25,11 @@ export function finishExperiment() {
     };
     window.addEventListener('click', removeDOMElements);
     window.addEventListener('keydown', removeDOMElements);
+    const logger = Logger.getInstance();
+    logger.capture('Task Finished', {
+      taskName: taskStore().task,
+      taskFinished: taskStore().taskComplete,
+    });
   }, 50); // delay so that previous key presses are not captured
 
   jsPsych.endExperiment(
--- a/task-launcher/src/tasks/shared/trials/finishExperiment.ts
+++ b/task-launcher/src/tasks/shared/trials/finishExperiment.ts
@@ -2,7 +2,6 @@ import { jsPsych } from '../../taskSetup';
 import { PageAudioHandler } from '../helpers';
 import { mediaAssets } from '../../..';
 import { taskStore } from '../../../taskStore';
-import { Logger } from '../../../utils';
 
 export function finishExperiment() {
   const t = taskStore().translations;
@@ -25,11 +24,6 @@ export function finishExperiment() {
     };
     window.addEventListener('click', removeDOMElements);
     window.addEventListener('keydown', removeDOMElements);
-    const logger = Logger.getInstance();
-    logger.capture('Task Aborted', {
-      taskName: taskStore().task,
-      taskFinished: taskStore().taskComplete,
-    });
   }, 50); // delay so that previous key presses are not captured
 
   jsPsych.endExperiment(
--- a/README.md
+++ b/README.md
@@ -98,3 +98,7 @@ External tasks:
 	- [ROAR Single Word Reading](https://roar.education/) [Language: Single-word Reading] - external?
 	- [ROAR Sentence Comprehension](https://roar.education/) [Language: Reading Comprehension] - external?
 	- Expressive Vocab (ROAR?)
+
+
+	NOTE:
+	“This project is tested with BrowserStack.”


task-launcher/src/tasks/shared/trials/finishExperiment.ts
```diff
@@ -26,7 +26,7 @@ export function finishExperiment() {
     window.addEventListener('click', removeDOMElements);
     window.addEventListener('keydown', removeDOMElements);
     const logger = Logger.getInstance();
-    logger.capture('Task Aborted', {
```
asengupta3, line 29: I think the name of the file is a misnomer. This function is called whenever we want to abort a task. The real task finished is here:
https://github.com/levante-framework/core-tasks/blob/main/task-launcher/src/tasks/shared/trials/taskFinished.ts#L44
@zwatson2001 anything to add/ correct?

zwatson2001:
I think it's just a naming issue - the finishExperiment file runs whenever the task ends early, which is usually because a participant got too many trials wrong or timed out. But these cases are still considered complete runs (they aren't 'aborted' in the sense that they got abandoned, the task just ended early by design). So maybe we should log "Task finished early" or something like that? Or if we don't want to distinguish failing out from finishing the task, we can just log  the same thing in either case. I can also rename the finishExperiment file, because it's definitely misleading.

PR #492, opened 2026-07-13
Log stopping rule on task finish

zwatson2001:
Updates the log messages based on how the task finished:

- User errored out: `Task finished: user reached max incorrect answers`
- User timed out: `Task finished: user reached max time`
- User completed all trials in the task timeline: `Task finished: user finished all trials`


digital-pro (approved):
This seems like a solid upgrade to our reporting on why tasks end early. LGTM.

PR #493, opened 2026-07-14
Matrix reasoning bug fixes 

zwatson2001:
This PR addresses Freshdesk ticket 89, fixing two related bugs in matrix reasoning:

- The "fall back" feature (which gives the younger kid instructions and items after a certain number of incorrect trials) was being triggered immediately after the beginning of the older kid block if the user got trials wrong earlier in the task. This will no longer happen for younger kids
- Incorrect answers in the younger kid block were causing the task to error out later as soon as the older kid block started

### DIFF
--- a/task-launcher/src/tasks/matrix-reasoning/timeline.ts
+++ b/task-launcher/src/tasks/matrix-reasoning/timeline.ts
@@ -247,7 +247,7 @@ export default function buildMatrixTimeline(config: Record<string, any>, mediaAs
       if (i % batchSize === 0) {
         preloadBatch();
       }
-      if (i <= fallbackIndex) {
+      if (i <= fallbackIndex && !heavyInstructions) {
         timeline.push(fallbackBlock);
       }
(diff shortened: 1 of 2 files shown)


u2234-io:
`downexInstructions4` is also part of `fallbackBlock` (L161), so older kids who fall back will now also get `numIncorrect` reset to 0, giving them a fresh `maxIncorrect` budget on resuming main trials. Is that intended behavior?

zwatson2001:
> `downexInstructions4` is also part of `fallbackBlock` (L161), so older kids who fall back will now also get `numIncorrect` reset to 0, giving them a fresh `maxIncorrect` budget on resuming main trials. Is that intended behavior?

Good catch - yes, I think the 'numIncorrect' reset should also apply to older kids that fell back (this is a change from the previous behavior, but it's consistent with preventing the user from erroring out immediately after receiving those instructions).

u2234-io (approved):
Gotcha. LGTM!

PR #495, opened 2026-07-15
SDS bugfixes

zwatson2001:
This PR fixes two bugs in SDS:

- The legacy CAT variant should no longer only display the "setup" trials in the something's the same phase
- Mashing the buttons on a touch screen in the test-dimensions phase should no longer prevent the task from advancing to the next trial


CCuskley (approved):
LGTM

PR #502, opened 2026-07-23
Minor improvements to buttons on touch screens

zwatson2001:
This PR adds changes that guard against button mashing in practice trials, which had prevented the task from advancing to the next trial. Practice buttons are now disabled until the feedback audio from an already-selected button is done playing. It also disables the hover property on touch screens, which prevents the buttons from appearing selected unexpectedly.


digital-pro (approved):
LGTM

PR #509, opened 2026-08-04
Use the same string keys for both audio and text in feedback trials

zwatson2001:
In memory game feedback trials, the top and bottom text was coming from two different string keys (for example in incorrect trials: `feedback-not-quite-right` and `memory-game-input`) but the audio was `memoryGameTryAgain`, which is a combination of the text in the separate individual text keys. This PR plays two separate audio files in the feedback trials to match the text keys used.

### DIFF
--- a/task-launcher/src/tasks/math/timeline.ts
+++ b/task-launcher/src/tasks/math/timeline.ts
@@ -94,7 +94,7 @@ export default function buildMathTimeline(config: Record<string, any>, mediaAsse
 
   const feedbackBlock = (trial?: StimulusType) => {
     return {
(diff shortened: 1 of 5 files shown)


u2234-io:
While testing this change I found that the new two-part feedback audio could bleed into the next trial: if you clicked Continue before the first audio finished, on_finish's stopAndDisconnectNode() fired the audio's onended, which then kicked off the chained prompt audio after the trial had already advanced. Added a trialFinished flag (reset in on_load, set in on_finish) so the chained prompt is skipped on early exit. Also camelized the audio key lookup so it stays consistent with the text lookup, and added a nullAudio fallback for the prompt. Verified on memory game (correct + incorrect), math, and SDS.

u2234-io (approved):

PR #512, opened 2026-08-06
fix: race on audio context resume

u2234-io:
For widest browser support, AudioContext `resume()` must only called synchronously within a user gesture handler. This was satisfied via handlers on `click` and `touchend` in the `enterFullscreen` trial.

However, `enterFullscreen` used `@jspsych/plugin-fullscreen` which set its own `click` handler on the same button element that would execute first, creating a race between `resume()` and subsequent audio autoplay. This would sometimes cause the first audio on the next trial to not play, which forced the user to replay it w/ the speaker button in order to advance.

I switched this trial to use `@jspsych-contrib/plugin-html-multi-response` in order to have control over when the trial advances in order to ensure `resume()` is called correctly and finishes before any audio plays in subsequent trials.

Deleted `ifNotFullscreen` trial - it was unused and would cause `resume()` to not get called if the user was already in fullscreen when starting the experiment.

Deleted the 1-frame silent buffer trick - webkit properly supported `resume()` ~2016 so no longer necessary.

### DIFF
--- a/task-launcher/src/tasks/shared/helpers/audioHandler.ts
+++ b/task-launcher/src/tasks/shared/helpers/audioHandler.ts
@@ -89,20 +89,4 @@ export class PageAudioHandler {
       return;
     }
   }
-
-  // required on iOS to prevent autoplay blocking
-  static unlockAudioContext() {
-    const ctx = jsPsych.pluginAPI.audioContext();
-
-    // safari requires resuming audio context on user interaction, then it can be used freely later
-    if (ctx.state === 'suspended') {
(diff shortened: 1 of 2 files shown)


u2234-io:
Note: trial type changed from `fullscreen` to `html-multi-response` in recorded data. If any downstream analysis filters on the `fullscreen` trial type, that data shape shifts. Probably irrelevant for an intro/enter-fullscreen trial, but flagging.

zwatson2001 (approved):

PR #513, opened 2026-08-06
fix: flower practice replay audio + prevent image dragging

zwatson2001:
Fixes freshdesk ticket https://levante-support.freshdesk.com/a/tickets/106. I wasn't able to replicate the problem described in this ticket: https://levante-support.freshdesk.com/a/tickets/105, but I disabled user dragging and touch callout, which will prevent images from being draggable and hopefully fix the issue mentioned in the ticket.


u2234-io (approved):
LGTM

